PR #194, opened 2014-11-02
doesn't it work also on python 3.4 ?

stonebig:
apparently yes on my windows 64bit  python3.4.2 pc, and yes for Travis on python 3.4.1 linux


AlJohri:
do you know if the current codebase also works with python 3.5? I'm working on a project that makes heavy use of asyncio and other python 3.5 features

dan-blanchard:
It should work fine with 3.5.

PR #227, opened 2015-02-11
Added KMeans unsupervised classifier

mklechan:
Would be nice to use unsupervised algorithms like KMeans. Of course there is no "evaluate" command to perform on this, but "train" would be fine.  Just want to output KMeans models within the skll framework.


dan-blanchard:
Thanks for the PR.  I like the idea of adding support for unsupervised algorithms to SKLL, but if we did that, we'd also need to:

-  Add checks to `experiments.py` to make sure people don't specify invalid tasks with unsupervised algorithms.
-  Add support for other unsupervised algorithms at the same time.
-  Add any algorithms that don't support dense to `skll.learner._REQUIRES_DENSE`.

desilinguist:
Sounds like this requires more work than, so I am going to close it for now.

Issue #222, opened 2015-02-04
Need cv_folds option


dan-blanchard:
I'm not sure how we missed this, but there's currently no `cv_folds` option exposed in the configuration files.  There's `cv_folds_files` and `grid_search_folds` (which apparently is undocumented, so we should fix that as well), but no access to `cv_folds` directly.

desilinguist:
Addressed by #240.

Issue #160, opened 2014-10-17
stratification during cross-validation should be optional 


mheilman:
The fact that SKLL uses stratification for cross-validation is not well-documented (it is only briefly mentioned in the Python API).  This should at least be documented.  It might be better to not use stratification by default.  It would also be good to make this an option somewhere.

desilinguist:
Addressed by #240.

PR #240, opened 2015-06-29
Expose cv_folds and stratified

aoifecahill:
This PR addresses #222 and #160. 

New options available in config file: 
``cv_folds``: specify the number of folds for cross validation
``random_folds``: when set to True, does not do stratified cross-validation

Added documentation for above options. 
Added missing documentation for ``grid_search_folds``. 

Fixed an apparent bug in ``experiments.py`` where the value of ``grid_search_folds`` set in the config file was being overwritten as 3 for model training when the task was not cross validation.

### DIFF
--- a/skll/experiments.py
+++ b/skll/experiments.py
@@ -405,7 +405,7 @@ def _parse_config_file(config_path):
         if num_cv_folds:
             try:
                 cv_folds = int(num_cv_folds)
-            except:
+            except ValueError:
                 raise ValueError("The value for cv_folds should be an integer. " +
                                  "You specified {}".format(num_cv_folds))
         else:
--- a/skll/experiments.py
+++ b/skll/experiments.py
@@ -246,6 +246,8 @@ def _setup_config_parser(config_path):
                                         'grid_search_jobs': '0',
                                         'grid_search_folds': '3',
                                         'cv_folds_file': '',
+                                        'num_cv_folds': '10',
+                                        'stratified_folds': 'True',
                                         'suffix': '',
                                         'label_col': 'y',
                                         'id_col': 'id',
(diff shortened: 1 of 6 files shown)


skll/experiments.py
```diff
@@ -387,13 +394,33 @@ def _parse_config_file(config_path):
...
     if cv_folds_file:
         cv_folds = _load_cv_folds(cv_folds_file,
                                   ids_to_floats=ids_to_floats)
     else:
-        cv_folds = 10
+        # set the number of folds for cross-validation
+        if num_cv_folds:
+            try:
+                cv_folds = int(num_cv_folds)
+            except:
```
desilinguist: It's probably better to specify a `ValueError` explicitly here.

skll/experiments.py
```diff
@@ -387,13 +394,33 @@ def _parse_config_file(config_path):
...
+        else:
+            # default number of cross-validation folds
+            cv_folds = 10
+
+    # whether or not to do stratified cross validation
+    random_folds = config.get("Input", "random_folds")
+    if random_folds == 'True':
+        if cv_folds_file:
+            logger.warning('Random folds will not override'+
+                           'values in cv_folds_file')
```
desilinguist: This message is not entirely clear. So, basically, random_folds is being ignored here, right? May be something like: `"Specifying cv_folds_file overrides random_folds"`?

skll/experiments.py
aoifecahill: I'm not sure what you mean? There is a ValueError raised?

desilinguist: Sorry, I meant that you want to catch a `ValueError` explicitly in the `except` clause.

aoifecahill: don't I do that?

desilinguist: Here's the code I see:

```python
        if num_cv_folds:
            try:
                cv_folds = int(num_cv_folds)
            except:
                raise ValueError("The value for cv_folds should be an integer. " +
                                 "You specified {}".format(num_cv_folds))
```

Here's what I am saying it should look like:

```python
        if num_cv_folds:
            try:
                cv_folds = int(num_cv_folds)
            except ValueError:
                raise ValueError("The value for cv_folds should be an integer. " +
                                 "You specified {}".format(num_cv_folds))
```

Note the difference in the `except` statement.

aoifecahill: Oh I see, will update.

Issue #238, opened 2015-05-28
Base estimators other than default are not supported for AdaBoost


benleong-ets:
Can we use SKLL's AdaBoost with LogisticRegression as the weak learner? We are interested to test out LogisticRegression instead of DecisionTreeClassifier.

In the config file, I set the base_estimator to be `LogisticRegression` as follows:

```
fixed_parameters=[{'base_estimator':'LogisticRegression', 'algorithm':'SAMME'}]
```

But I got the following error:

```python
TypeError: Cannot clone object ''LogisticRegression'' (type <class 'str'>): it does not seem to be a scikit-learn estimator it does not implement a 'get_params' methods.
```

The input param is specified as a string but an object is needed. We can't do the conversion with the config file.

desilinguist:
Addressed by #245.

Issue #234, opened 2015-04-15
Bug when checking if `hasher_features` is a valid option


desilinguist:
In `experiments.py`, we have the following code:

```python
# produce warnings if hasher_features is set but feature_hasher
# is missing or set to False.
if config.has_option("Input", "hasher_features"):
    if not config.has_option("Input", "hasher_features"):
        logger.warning("Ignoring hasher_features since feature_hasher" +
                       " is missing from the config file.")
    else:
        feature_hasher = config.getboolean("Input", "feature_hasher")
        if not feature_hasher:
            logger.warning("Ignoring hasher_features since feature_hasher" +
                           " is set to False.")
```

The second `if` statement should read:
```python
    if not config.has_option("Input", "feature_hasher"):
```

dan-blanchard:
Man, who would have done such a thing? (7a434e2173a) :stuck_out_tongue_winking_eye:

desilinguist:
I figured it was a copy-paste flub and I am the master of those :)

dan-blanchard:
Yeah, and it doesn't help how similar those two settings names are.

desilinguist:
Addressed by #245.

Issue #220, opened 2014-12-10
Stop modifying global numpy random seed


mheilman:
In `learner.train` and `learner.cross_validate`, the global numpy random seed is set.  This seems unnecessary since the randomized algorithms in scikit-learn generally take random seed arguments.  Can we remove these?

https://github.com/EducationalTestingService/skll/blob/5c113bca947082e2b56e41c8a13d7ab43d49160a/skll/learner.py#L907
https://github.com/EducationalTestingService/skll/blob/5c113bca947082e2b56e41c8a13d7ab43d49160a/skll/learner.py#L1393

desilinguist:
Addressed by #245.

Issue #207, opened 2014-11-21
Remove decision tree and random forest learners from requires_dense set


dan-blanchard:
Whenever the next version of scikit-learn comes out, it will support sparse input for decision trees and random forests, which is pretty great news.  This was added in scikit-learn/scikit-learn#3173.

Anyway, we'll need to update our `skll.learner._REQUIRES_DENSE` set to remove those learners from it to take advantage of this.

dan-blanchard:
0.16 is out with this, so we should do it.

desilinguist:
Addressed by #245.

PR #245, opened 2015-07-14
Several minor bugfixes

desilinguist:
- Allow non-default base estimators for AdaBoost (#238)
  - Correctly interpret the `base_estimator` fixed parameter for AdaBoostClassifiers and AdaBoostRegressors as objects instead of strings.
  - Make sure that the base estimator for Adaboost has a fixed seed so that results are replicable.
  - Add corresponding unit tests for both classification and regression and update documentation.
- Fix `hasher_features` typo in `experiments.py` (#234).
- Remove unnecessary setting of numpy global random seeds (#220)
  - We are already passing in `random_state` seeds in `model_kwargs` for all classifiers/regressors so we don't need to set the `numpy` global random seed.
  - Include `Lasso`, `ElasticNet`, `LinearSVR`, and `SVC` when setting the above seeds in `model_kwargs`.
- Remove requirements for several learners to have dense input (#207)
  - In the new version of scikit-learn, only `GradientBoostingClassifier` and `GradientBoostingRegressor` require dense input.
  - Move `make_sparse_data()` to `utils.py`
  - Update `test_sparse_predict()` to include all sparse-friendly classifiers.
- Update expected values in certain tests to account for all of the above changes.

### DIFF
--- a/doc/run_experiment.rst
+++ b/doc/run_experiment.rst
@@ -145,7 +145,7 @@ possible settings for each section is provided below, but to summarize:
 .. _evaluate:
 
 *   If you want to **train a model and evaluate it** on some data, specify a
-    training location, a test location, and a directory to store results, 
+    training location, a test location, and a directory to store results,
     and set :ref:`task` to ``evaluate``.
 
 .. _predict:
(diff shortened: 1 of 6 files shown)


dan-blanchard:
:+1: I looked at this on my phone at lunch and forgot to comment.  Looks good to me.

aoifecahill:
Looks good to me too, thanks!

Issue #223, opened 2015-02-04
need to improve the way to check the validity of configuration file


cml54:
In the current stable version, when a user sets a configuration name in a wrong section, skll can't detect it. For example, 'cv_folds_file' should be under 'Input' section. When a user sets it under 'General' section, skll recognizes the configuration in a correct format. It needs to be improved.

desilinguist:
Addressed by #246.

Issue #213, opened 2014-11-23
Config file paths are relative to current working directory instead of config file location


brianray:
These may be windows related. However, two things I ran into so far one small and the other will need more investigation.

a) The path issue

```pytb
C:\Users\brray\Documents\GitHub\skll\examples>run_experiment titanic/evaluate_tuned.cfg
Traceback (most recent call last):
  File "C:\Users\brray\AppData\Local\Continuum\Anaconda\lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Users\brray\AppData\Local\Continuum\Anaconda\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Users\brray\AppData\Local\Continuum\Anaconda\Scripts\run_experiment.exe\__main__.py", line 9, in <module>
  File "C:\Users\brray\AppData\Local\Continuum\Anaconda\lib\site-packages\skll\utilities\run_experiment.py", line 108, in main
    ablation=ablation, resume=args.resume)
  File "C:\Users\brray\AppData\Local\Continuum\Anaconda\lib\site-packages\skll\experiments.py", line 1052, in run_configuration
    custom_learner_path) = _parse_config_file(config_file)
  File "C:\Users\brray\AppData\Local\Continuum\Anaconda\lib\site-packages\skll\experiments.py", line 459, in _parse_config_file
    "config file does not exist"), train_path)
IOError: [Errno 2] The training path specified in the config file does not exist: u'train'
```

If I 'cd' into the titanic directory, it works.

desilinguist:
Addressed by #246.

PR #246, opened 2015-07-14
Overhaul configuration file parsing

desilinguist:
- Use a new custom parser class `SKLLConfigParser` based on `ConfigParser`.
- Move all of the configuration parsing code to `config.py` and out of `experiments.py`.
- Add a `validate()` method to the config parser that raises a `KeyError` if:
  - unrecognized options are specified
  - options are specified in more than one section
  - options are specified in the wrong section (#223)
- Add unit tests for the above validation checks
- Update tests to deal with the above changes
- Update `requirements_rtd.txt` to install the latest version of `ConfigParser` (v3.5.0b2)


desilinguist:
@dan-blanchard please do look at this one too since it's a somewhat major change.

tests/test_input.py
```diff
@@ -671,16 +643,91 @@ def test_config_parsing_bad_cv_folds():
...
+
+    yield check_config_parsing_key_error, config_path
+
+
+def test_config_parsing_duplicate_option():
+
+    train_dir = join(_my_dir, 'train')
+    output_dir = join(_my_dir, 'output')
+
+    # make a simple config file that has an invalid option
```
aoifecahill: minor: comment should be updated to reflect that the problem with this config is that there is a duplicate key (that comes from the fill_in_config_options method)

aoifecahill:
This looks good to me, I didn't spot anything obvious that might cause problems.

dan-blanchard:
I did a little PEP8 clean up here and there, but otherwise this looks great! :+1:

desilinguist:
Thanks, Dan! I will merge as soon as the build passes.

dan-blanchard:
I actually already fixed that comment too. :)

desilinguist:
Sweet! Thanks for that!

Actually, I just realized I didn't address #213 in this PR. Do you think that's important for this new release?

dan-blanchard:
It would be nice to have fixed for this release, since we know it's bitten some new users when trying to follow the tutorial.

desilinguist:
Okay, I will add that to this PR then and try to add some tests for it too.

desilinguist:
@dan-blanchard and @aoifecahill, can you please take a look again? I just committed some changes to address #213.

skll/config.py
```diff
@@ -0,0 +1,575 @@
...
+    if random_folds:
+        if cv_folds_file:
+            logger.warning('Specifying cv_folds_file overrides random_folds')
+        do_stratified_folds = False
+    else:
+        do_stratified_folds = True
+
+    # get all the input paths and directories (without trailing slashes)
+    train_path = config.get("Input", "train_directory").rstrip('/')
+    test_path = config.get("Input", "test_directory").rstrip('/')
```
dan-blanchard: This won't work on Windows.

desilinguist: This has been in there since the beginning so I guess it has never worked?

desilinguist:
I replaced the `/` with `os.sep` which should work on Windows too.

tests/test_input.py
```diff
@@ -671,16 +645,151 @@ def test_config_parsing_bad_cv_folds():
...
-                                                   'bad_cv_folds')
+                                                   'mislocated_input_file')
+
+    yield check_config_parsing_file_not_found_error, config_path
 
-    _parse_config_file(config_path)
+
+def test_config_parsing_relative_input_path():
+
+    train_dir = '../train'
```
dan-blanchard: This should also be:

```python
train_dir = join('..', 'train')
```

desilinguist: Ah, good call.

desilinguist:
So, is this okay to be merged once I address your last comment?

dan-blanchard:
Definitely.

Issue #250, opened 2015-08-17
print_model_weights does not work for Logistic Regression models without intercepts.


desilinguist:
Here's the error I get:

```python
Traceback (most recent call last):
  File "/Users/nmadnani/anaconda/envs/garbage/bin/print_model_weights", line 6, in <module>
    sys.exit(main())
  File "/Users/nmadnani/anaconda/envs/garbage/lib/python3.4/site-packages/skll/utilities/print_model_weights.py", line 55, in main
    (weights, intercept) = learner.model_params
  File "/Users/nmadnani/anaconda/envs/garbage/lib/python3.4/site-packages/skll/learner.py", line 712, in model_params
    intercept = dict(zip(label_list, self.model.intercept_))
TypeError: zip argument #2 must support iteration
```

dan-blanchard:
Fixed by #251.

PR #251, opened 2015-08-26
Fix some bugs relating to intercept

aoifecahill:
Stops a crash in learner.py when there is no intercept. 

Stops a crash in print_model_weights.py when the intercept is an array.

### DIFF
--- a/skll/utilities/print_model_weights.py
+++ b/skll/utilities/print_model_weights.py
@@ -65,7 +65,7 @@ def main(argv=None):
         # subclass of LinearModel
         if '_intercept_' in intercept:
             # Some learners (e.g. LinearSVR) may return a list of intercepts
-            if type(intercept['_intercept_']) == np.ndarray:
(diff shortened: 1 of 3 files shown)


skll/utilities/print_model_weights.py
```diff
@@ -63,7 +64,12 @@ def main(argv=None):
     if intercept is not None:
         # subclass of LinearModel
         if '_intercept_' in intercept:
-            print("intercept = {:.12f}".format(intercept['_intercept_']))
+            # Some learners (e.g. LinearSVR) may return a list of intercepts
+            if type(intercept['_intercept_']) == np.ndarray:
```
dan-blanchard: This needs to be:

```python
if isinstance(intercept['_intercept_'], np.ndarray):
```

otherwise it won't work on subclasses.

aoifecahill: Ah, thanks.

dan-blanchard:
This fixes #250.

PR #254, opened 2015-09-08
correct total number of examples

aoifecahill:
Fix the division by zero error by adding one to the enumerator counter `ex_num`


dan-blanchard:
Since there are two spots where we're adding one to `ex_num`, it would probably be a tiny bit cleaner to just add `start=1` to the `enumerate()` calls.

PR #260, opened 2015-10-22
Fix issue where os.makedirs was being called for directories that already existed

dan-blanchard:
I got an email from a user about getting a `FileExistsError` when running the titanic example, and I noticed that we were checking to see if one path existed and then called `os.makedirs` to make a slightly different one existed.  This should fix that.

### DIFF
--- a/skll/config.py
+++ b/skll/config.py
@@ -194,14 +194,14 @@ class SKLLConfigParser(configparser.ConfigParser):
                                                   incorrectly_specified_options]))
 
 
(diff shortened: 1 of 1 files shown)


desilinguist:
Ah, good catch.

dan-blanchard:
Without this fix many config files would not work properly, so we should probably release a bugfix release.

desilinguist:
Sounds good to me.

PR #262, opened 2015-10-22
Feature/save cv folds

aoifecahill:
Add a new config option that saves information about which example was in which test fold during cross validation experiments. The name of the file is fixed for each experiment.

### DIFF
--- a/skll/config.py
+++ b/skll/config.py
@@ -70,6 +70,7 @@ class SKLLConfigParser(configparser.ConfigParser):
                     'results': '',
                     'sampler': '',
                     'sampler_parameters': '[]',
+                    'save_cv_folds': 'False',
                     'shuffle': 'False',
                     'suffix': '',
                     'test_directory': '',
@@ -105,6 +106,7 @@ class SKLLConfigParser(configparser.ConfigParser):
                                    'results': 'Output',
                                    'sampler': 'Input',
                                    'sampler_parameters': 'Input',
+                                   'save_cv_folds': 'Input',
                                    'shuffle': 'Input',
                                    'suffix': 'Input',
                                    'test_directory': 'Input',
@@ -363,6 +365,9 @@ def _parse_config_file(config_path):
             # default number of cross-validation folds
             cv_folds = 10
 
+    # whether or not to save the cv fold ids
+    save_cv_folds = config.get("Input", "save_cv_folds")
+
     # whether or not to do stratified cross validation
     random_folds = config.getboolean("Input", "random_folds")
     if random_folds:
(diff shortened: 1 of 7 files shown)


tests/test_cv.py
```diff
@@ -43,6 +63,20 @@ def tearDown():
     if exists(fold_file_path):
         os.unlink(fold_file_path)
 
+    train_dir = join(_my_dir, 'train')
+    output_dir = join(_my_dir, 'output')
+    config_dir = join(_my_dir, 'configs')
+
+    cfg_file = join(config_dir, 'test_save_cv_folds.cfg')
+    os.unlink(cfg_file)
```
mulhod: Just wondering: Is there some specific reason we're using unlink here as opposed to remove or is it just because that's what has been used in the past? The docs say that it's identical to remove and it seems like the naming is clearer there.

desilinguist: It's just a consistency thing. We have been using `unlink` in the pat.

tests/test_cv.py
```diff
@@ -177,3 +211,67 @@ def test_load_cv_folds_non_float_ids():
...
+def test_retrieve_cv_folds():
+    """
+    Test to make sure that the fold ids get returned correctly after cross-validation
+    """
+
+    learner = Learner('LogisticRegression')
+    num_folds = 5
+    cv_fs, custom_cv_folds = make_cv_folds_data(num_examples_per_fold=2, num_folds=num_folds)
+
+    expected_fold_ids = {'EXAMPLE_0': 0, 'EXAMPLE_1': 4, 'EXAMPLE_2': 3, 'EXAMPLE_3': 1,
```
desilinguist: Isn't this the same as `custom_cv_folds`? Would it make more sense to use that as the expected output since that was what was used to create the input folds in the first place?

aoifecahill: But that's what I'm testing here, that ``custom_cv_folds`` is actually being correctly returned from the function that computes it. I'm misunderstanding something?

aoifecahill: I see my confusion. This ``learner.cross_validate`` call was not using those ``custom_cv_folds``. Updating tests to account for both cases: (1) where the user passes in custom folds and (2) where skll generates them.

skll/config.py
```diff
@@ -363,6 +365,9 @@ def _parse_config_file(config_path):
             # default number of cross-validation folds
             cv_folds = 10
 
+    # whether or not to save the cv fold ids
+    save_cv_folds = config.get("Input", "save_cv_folds")
```
dan-blanchard: I think this belongs in `Output`, since this an option that relates to what we're outputting, and not to what we're given.

desilinguist:
This looks pretty good to me. I am going to merge it tomorrow unless anyone else has an objection.

PR #264, opened 2015-10-22
Bugfix/update config.py

mulhod:
Fix minor errors that I noticed while going through config.py.

### DIFF
--- a/skll/config.py
+++ b/skll/config.py
@@ -349,22 +349,14 @@ def _parse_config_file(config_path):
 
     # get the cv folds file and make a dictionary from it, if it exists
     cv_folds_file = config.get("Input", "cv_folds_file")
-    num_cv_folds = config.get("Input", "num_cv_folds")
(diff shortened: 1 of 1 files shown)


mulhod:
I don't really know why this is failing. The errors don't seem to be in any way related to the PR, which would be very unexpected anyway. But other recent PRs haven't failed in a similar way, so I don't really know.

dan-blanchard:
Binstar went down yesterday for a bit, so that's why it was failing. I restarted it.

mulhod:
Any chance we can try to do the build again for this one again? Is there any reason to think that it's failing due to what's in the PR?

dan-blanchard:
The issue was that they changed the default path for where miniconda gets installed, which broke all our travis builds.  I fixed that in master and then rebased your branch, so everything went fine after that.

mulhod:
Thanks.

PR #266, opened 2015-10-26
only print out a warning about hasher_features when feature_hasher is not set

### LINKED ISSUE #265: feature_hasher check in config.py prints out misleading warnings.
It is always ignored.

aoifecahill:
Otherwise the user might think that their feature hasher is being ignored. 

This fixes #265.

### DIFF
--- a/skll/config.py
+++ b/skll/config.py
@@ -276,7 +276,7 @@ def _parse_config_file(config_path):
 
     # produce warnings if hasher_features is set but feature_hasher
     # is not set correctly
-    if not feature_hasher and hasher_features > 0:
+    elif hasher_features > 0:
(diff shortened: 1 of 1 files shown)


skll/config.py
```diff
@@ -276,7 +276,7 @@ def _parse_config_file(config_path):
 
     # produce warnings if hasher_features is set but feature_hasher
     # is not set correctly
-    if hasher_features > 0:
+    if not feature_hasher and hasher_features > 0:
```
dan-blanchard: Good catch, but this could be simplified to:

```py
elif hasher_features > 0:
```

because we're checking if `feature_hasher` is set right about this.  I imagine that was the original intention for this, but there was a typo.

aoifecahill: Good point, thanks!

PR #267, opened 2015-11-09
Fix a bug in print_model_weights for multi class classification

aoifecahill:
Also added a test for this case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
   - sudo chmod 777 /scratch/
   - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
   - chmod +x miniconda.sh
-  - ./miniconda.sh -b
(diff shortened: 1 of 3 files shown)


desilinguist:
Looks like this PR fails on the gridmap-enabled builds because `drmaa` throws the following error:

```python
InternalException: code 1: reresolve hostname failed: hostname exeeds hostname length(MAXHOSTNAMELEN) on this system
```

@dan-blanchard, any ideas?

dan-blanchard:
It's a bug with Travis.  We could [add this workaround](https://github.com/travis-ci/travis-ci/issues/5227#issuecomment-165131913) to our `.travis.yml`, or just wait for them to fix it (since it looks like they've been working on it in the last 24 hours).  Relevant issues: travis-ci/travis-ci#5292 and travis-ci/travis-ci#5227.

desilinguist:
I think we can wait another day or two.

desilinguist:
I think we might need to apply the workaround now. It's been too long and they haven't fixed it.

PR #276, opened 2016-01-19
Workaround for travis-ci/travis-ci#5227

desilinguist:
Since Travis still hasn't gotten around to the "non-invasive" solution, we need to try the workaround.


desilinguist:
@dan-blanchard seems like it's now failing with a different grid engine error about not being able to find any suitable queues. Thoughts?

.travis.yml
```diff
@@ -10,6 +10,12 @@ notifications:
 
 # Install stuff
 before_install:
+  - hostname
+  - cat /etc/hosts # optionally check the content *before*
+  - sudo hostname "$(hostname | cut -c1-63)"
+  - sed -e "s/^\\(127\\.0\\.0\\.1.*\\)/\\1 $(hostname | cut -c1-63)/" /etc/hosts | sudo tee /etc/hosts
```
dan-blanchard: You need to update line 8 in `install_sge.sh` to account for this change.  It also modifies `/etc/hosts` and it won't be able to find the pattern it's looking for until you update that.

desilinguist:
I am a little confused. Line 8 of `install_sge.sh` seems written to convert this line:

```
127.0.0.1 localhost.localdomain localhost
```

into this line:

```
127.0.0.1 localhost localhost.localdomain <hostname>.
```

where `<hostname>` is the hostname of the machine.

However, `/etc/hosts` before I do anything else looks like this:

```
127.0.0.1       localhost
127.0.1.1       precise64
# host on which base box was built. Erlang services that use Mnesia
# (like RabbitMQ) may create DB directories with this hostname in the name
# during provisioning and if we change hostname later, they will have all
# kinds of obscure issues. MK.
127.0.0.1       nettuno
# backwards compatibility. MK.
127.0.0.1       travis
127.0.0.1       vagrant
```

So, the `install_sge.sh` regular expression won't match anyway right? Am I missing something?

dan-blanchard:
Oh wow, you're right.  It seems they changed two things on us when they decided to make the hostnames really long.  So yeah, take out that bit about `localhost.localdomain`.

desilinguist:
Actually, do we really need to do anything with `/etc/hosts` in `install_sge.sh` at all? Isn't this exactly what the workaround `sed` command: 

```bash
sed -e "s/^\\(127\\.0\\.0\\.1.*\\)/\\1 $(hostname | cut -c1-63)/" /etc/hosts | sudo tee /etc/hosts
```

is doing anyway?

desilinguist:
@dan-blanchard if you have the time, can you please look at the logs and give me a hint as to how to fix this? Otherwise, I am planning to turn off the two SGE travis builds since we really need to release the next version of SKLL in the next few weeks.

dan-blanchard:
@desilinguist Unfortunately, I don't have a lot of time to dedicate to this at the moment.  [This old post](http://dan-blanchard.roughdraft.io/6586533-how-to-setup-a-single-machine-sun-grid-engine-installation-for-unit-tests-on-tr) I put together explains the logic behind the SGE installation process.  I think there are two problems here, but I'm not sure what exactly to do to solve them:

1.  The default hostname is too long
2.  The hostname and localhost must point to the same IP

It seems like what you have already in this PR should fix both of them, but it still doesn't work.  You can see the cluster isn't setup properly when you look at the output from `install_sge.sh` and see:

```
Printing queue info to verify that things are working correctly.
queuename                      qtype resv/used/tot. load_avg arch          states
---------------------------------------------------------------------------------
all.q@localhost                BIP   0/0/2          -NA-     -NA-          au
	error: no value for "np_load_avg" because execd is in unknown state
You should see sge_execd and sge_qmaster running below:
travis    1985  0.0  0.0  11064  1472 pts/0    S+   19:53   0:00 /bin/bash travis/install_sge.sh
sgeadmin  3975  0.3  0.0 140284  5564 ?        Sl   19:54   0:00 /usr/lib/gridengine/sge_qmaster
root      5054  0.0  0.0  56096  3216 ?        Sl   19:54   0:00 /usr/lib/gridengine/sge_execd
travis    5100  0.0  0.0   8112   928 pts/0    S+   19:54   0:00 grep sge
```

Maybe `/etc/hosts/` workaround isn't enough to change the actual hostname to be short enough to not break `sge_execd`.  That's about the only thing I can think of that could be wrong at this point.

desilinguist:
I even tried setting up a local vagrant Ubuntu 12.04 box with the hostname in the Vagrantfile set to be the same as what travis seems to be setting it to. The scripts work just fine and `qstat` does not output any errors. 

At this point, the lack of SSH access in Travis seems to be a big gaping hole. Both CircleCI and Codeship offer that. The latter also offers unlimited builds for open-source projects with their free plan. I think we should just abandon Travis CI at this point and move to one of the other CI platforms. Or we just don't test the `gridmap` enabled builds at all.

dan-blanchard:
Fair enough. The Travis folks do seem overly willing to break complex
builds, so please feel free to switch us over to Codeship. I remember
looking at it before and thinking it was pretty nice.

desilinguist:
So, I looked into 3 other CI platforms: 

1. **CircleCI**: Looks like they are having trouble with running builds on multiple python versions. See discussion [here](https://discuss.circleci.com/t/testing-in-different-environments/450/7).

2. **Codeship**: I don't even think they support a build matrix (multiple python versions x other build environment variables) in a straightforward way. I have sent them an email asking if they do.

3. **Shippable**: Seems like the option that is most similar to Travis CI and they use docker containers for everything. They do support the build matrix we want but they only give you one "minion" at a time - which means our 4 builds will run sequentially. This is likely to take 4 hours.

I was thinking about it. Why can't we just use Bamboo that we have in-house at ETS? It supports GitHub repositories and we won't even need to install SGE. We can run all four builds in parallel too. And with this [plugin](https://marketplace.atlassian.com/plugins/com.mhackner.bamboo-github-status/server/overview), we can even show build notifications on GitHub. Thoughts?

dan-blanchard:
That sounds reasonable, but unless things have changed a lot with Bamboo recently, there are probably some security considerations that would need to taken into account before doing that.  Namely, you're going to want to be building things in containers or chroot jails so that some jackass can't submit a PR that has a malicious test in it that will get run automatically on the ETS servers.

desilinguist:
Good point. I was already planning to set it up such that no builds run automatically though. We will run things manually after we are satisfied that it's worth running. This is important because ETS-internal jobs should always take priority over SKLL jobs anyway. We will certainly lose some of the automation that we have with Travis CI but I think the benefits are worth it.

dan-blanchard:
Outside contributors not being able to see if their builds pass automatically is kind of a big deal to me (as an outside contributor).  I don't think making it manual makes a lot of sense.  I think the lesser evil here is just disabling the gridmap jobs on Travis temporarily.

desilinguist:
I just copied `.travis.yml` to `shippable.yml` and it seems to be working. I need to tweak some scripts to make them less travis-centric but I think this is a good option. I will work on it.

dan-blanchard:
Cool, thanks for handling this mess!

desilinguist:
Well, Shippable also does not allow SSH access, looks like and SGE installation seems to be failing there for a different reason I can't really debug. I am going to revert back to using Travis and turn off the gridmap builds.

PR #278, opened 2016-01-27
Switch to using new travis infrastructure.

desilinguist:
- Remove all traces of SGE installation and log capturing.
- Set `sudo: false` in `.travis.yml` to use Travis' new container infrastructure.


dan-blanchard:
Seems reasonable to me.  Thank you for making this a single commit so that we can revert it easily when they fix this stuff.

PR #281, opened 2016-02-12
Raise error when config file on command line is an empty string.

aloukina:
Added an extra check for empty config file name. This can generally only happen via the API.

### DIFF
--- a/skll/config.py
+++ b/skll/config.py
@@ -236,6 +236,10 @@ def _parse_config_file(config_path):
     # Initialize logger
     logger = logging.getLogger(__name__)
 
+    # check that config_path is not empty
+    if config_path == "":
+        raise ValueError("The name of the configuration file is empty")
+
     # compute the absolute path for the config file
     config_path = realpath(config_path)
     config_dir = dirname(config_path)
--- a/skll/config.py
+++ b/skll/config.py
@@ -238,7 +238,7 @@ def _parse_config_file(config_path):
 
     # check that config_path is not empty
     if config_path == "":
-        raise ValueError("The name of the configuration file is empty")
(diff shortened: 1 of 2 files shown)


skll/config.py
```diff
@@ -236,6 +236,10 @@ def _parse_config_file(config_path):
     # Initialize logger
     logger = logging.getLogger(__name__)
 
+    # check that config_path is not empty
+    if config_path == "":
+        raise ValueError("The name of the configuration file is empty")
```
aoifecahill: perhaps FileNotFoundError would be more appropriate?

aoifecahill:
Adding more scope to this issue: what happens when the config file does not exists (different from it being blank)?

Could you also account for that case and add the relevant test?

aloukina:
This is already addressed - SKLL raises IOError. The relevant test is in test_input line 130.

aoifecahill:
Then maybe we should raise IOError here too?

aoifecahill:
actually, if we are already checking for a file that doesn't exist, then I think this check belongs in the same place (``_setup_config_parse``)

aloukina:
I thought of this but by that point the config_path is already converted to absolute path by realpath(). In case of empty filename this would be just the directory name which exists. So to catch the error we'll need to be checking whether supplied config_path is a directory which is a less transparent check.

aoifecahill:
fair enough :)

desilinguist:
So, you just need to change the `ValueError` to an `IOError`, sounds like?

desilinguist:
Since, `coveralls.io` is still down. I ran this manually and the coverage was 91%. So, I am going to go ahead and merge this.

PR #286, opened 2016-02-12
Pass fixed random state to kfold

bwriordan:


tests/test_cv.py
```diff
@@ -222,20 +222,26 @@ def test_retrieve_cv_folds():
...
     _, _, skll_fold_ids = learner.cross_validate(cv_fs, cv_folds=custom_cv_folds, save_cv_folds=True,
                                                  grid_search=True)
+    assert_equal(skll_fold_ids, custom_cv_folds)
 
+    # Test when learner.cross_validate() makes the folds but stratified=False and grid_search=False.
+    expected_fold_ids = {'EXAMPLE_0': '0', 'EXAMPLE_1': '0', 'EXAMPLE_2': '1', 'EXAMPLE_3': '1',
+                         'EXAMPLE_4': '2', 'EXAMPLE_5': '2', 'EXAMPLE_6': '3', 'EXAMPLE_7': '3',
+                         'EXAMPLE_8': '4', 'EXAMPLE_9': '4'}
+    _, _, skll_fold_ids = learner.cross_validate(cv_fs, cv_folds=custom_cv_folds, save_cv_folds=True,
+                                                 stratified=False, grid_search=False)
```
desilinguist: I am confused. The comment says that we are letting `learner.cross_validate()` make the folds itself but then we are also passing in `custom_cv_folds`?

bwriordan:
Thanks for catching. The reason that test was even passing was that the make_cv_folds_data() function has a random number generator with the same seed as is now set at the top of learner.cross_validate(), so expected and actual fold ids matched.

aoifecahill:
Looks good to me.

desilinguist:
@bwriordan can you please merge in the latest `master`?

PR #287, opened 2016-02-12
Fix errors when using reusing relative paths in output section

mulhod:
Added in support of relative paths in config files for more than just the training/test set files (e.g. model directory, custom learner, etc.) + some unit tests.


desilinguist:
This looks really good! Only thing - you should probably modify `the `tearDown()` method in `test_input.py` to also delete the instantiated config file.

desilinguist:
Thanks! Looks good to me.

aoifecahill:
Looks good to me.

PR #288, opened 2016-02-12
Update scikit-learn to 0.17.0

desilinguist:
The only change that was required was to patch up learner objects from older versions of SKLL (and scikit-learn) that we might load using the `from_file()` class method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
   - conda config --add channels https://conda.binstar.org/dan_blanchard
   - conda update --yes conda
 install:
-  - conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy beautiful-soup six scikit-learn==0.16.1 joblib prettytable python-coveralls pyyaml
+  - conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy beautiful-soup six scikit-learn==0.17.0 joblib prettytable python-coveralls pyyaml
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes configparser logutils mock; fi
   # Have to use pip for nose-cov because its entry points are not supported by conda yet
   - pip install nose-cov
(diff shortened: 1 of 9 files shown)


desilinguist:
I also ran the unit tests with `gridmap` on python 2.7 and 3.3. Everything passes.

dan-blanchard:
:+1: Looks good to merge to me.

desilinguist:
I looked through the [release notes](http://scikit-learn.org/stable/whats_new.html) for scikit-learn 0.17.0 and made any appropriate changes. But please take a look to see if I have missed anything.

dan-blanchard:
We should verify that we don't pass 1D arrays anywhere:

> Passing 1D data arrays as input to estimators is now deprecated as it caused confusion in how the array elements should be interpreted as features or as samples. All data arrays are now expected to be explicitly shaped (n_samples, n_features). By Vighnesh Birodkar.

We should rely on this new attribute instead of doing the subclass check we do now:

> Classifier and Regressor models are now tagged as such using the `_estimator_type` attribute.

I'd have to take a closer look if this is a problem or not for us:

> Cross-validation iterators allways provide indices into training and test set, not boolean masks.

Those are the only things that jumped out at me.

desilinguist:
Thanks @dan-blanchard!

1. I don't see any 1D arrays being passed to any `.fit()` or `.transform()` methods for estimators.

2. I replaced all the `issubclass(..., RegressorMixin)` to check for `_estimator_type` instead.

3. We have three cross-validation iterators: `LeaveOneLabelOut`, `StratifiedKFold`, and `KFold. All of them are used in `learner.py` and all of them are being used as iterators over indices.

AlJohri:
Can you push a minor version upgrade with this change if 1.2 is far off? Looking to try skll in my pipeline. Installing HEAD from git for now. Thanks!

`pip install "git+git://github.com/EducationalTestingService/skll.git#egg=skll"`

desilinguist:
@AlJohri actually we are planning to release 1.2 early next week :)

AlJohri:
:+1:

PR #291, opened 2016-02-12
Allow multiple objective functions in configuration files

bndgyawali:
Now, objectives replaced by objective. However, if any config file contains objective, it is normalized to objectives. objectives should be a list of objective functions


skll/config.py
```diff
@@ -485,11 +485,20 @@ def _parse_config_file(config_path):
...
 
-    # what is the objective function for the grid search?
-    grid_objective = config.get("Tuning", "objective")
-    if grid_objective not in SCORERS:
-        raise ValueError('Invalid grid objective function: {}'
-                         .format(grid_objective))
+    # what are the objective functions for the grid search?
+    grid_objectives_string = config.get("Tuning", "objective")
+    grid_objectives = yaml.load(_fix_json(grid_objectives_string))
+    if isinstance(grid_objectives, str):
```
desilinguist: In other places in the code, we usually do `isinstance(..., string_types)` so we should probably do the same here.

skll/config.py
```diff
@@ -485,11 +485,20 @@ def _parse_config_file(config_path):
...
+        grid_objectives_list = [grid_objectives]
+    elif isinstance(grid_objectives, list):
+        grid_objectives_list = grid_objectives
+    else:
+        raise ValueError("objective should be a string or "
+                         "list of objectives")
+
+    if not set(grid_objectives_list).issubset(set(SCORERS.keys())):
```
desilinguist: Why do we want a `set` here? Are we expecting people to use the same objective function twice?

skll/config.py
```diff
@@ -485,11 +485,20 @@ def _parse_config_file(config_path):
...
+        raise ValueError('Invalid grid objective function/s: {}'
+                         .format(list(grid_objectives)))
```
desilinguist: You can just use `grid_objectives_list` here instead of creating a list again from `grid_objectives`?

tests/configs/test_ablation.template.cfg
```diff
@@ -9,7 +9,7 @@ suffix=.jsonlines
 
 [Tuning]
 grid_search=False
-objective=accuracy
+objective=["accuracy"]
```
desilinguist: Why are you changing the templates if the code works fine with a string? There's no reason to.

desilinguist:
This looks pretty good aside from a few minor comments that can be easily addressed! 

Thanks again for helping with this.

skll/config.py
```diff
@@ -485,11 +485,18 @@ def _parse_config_file(config_path):
...
+    # what are the objective functions for the grid search?
+    grid_objectives_string = config.get("Tuning", "objective")
+    grid_objectives = yaml.load(_fix_json(grid_objectives_string))
+    if isinstance(grid_objectives, string_types):
+        grid_objectives = [grid_objectives]
+    elif not isinstance(grid_objectives, list):
+        raise ValueError("objective should be a string or "
+                         "list of objectives")
+
+    if not all([(True if objective in SCORERS else False) for objective in grid_objectives]):
```
desilinguist: Do we really need to be explicit with `True` and `False`? Can't you just say `if not all([objective in SCORES for objective in grid_objectives])' ?

desilinguist:
Now that I think about it, this PR is likely going to break anything that consumes SKLL `results` or `predictions` files since we are essentially changing their names by adding the name of the metric into it. 

So, perhaps, we should err on the side of backwards compatibility, and only add the name of the metric in the names of those files if we have more than one metric? @aoifecahill thoughts?

aoifecahill:
Good point. Let's try and maintain backward compatibility as much as possible.

desilinguist:
I agree. Sorry, @bndgyawali, more work for you :/

bndgyawali:
@desilinguist  Would you please review this again. This commit reverts most of the previous changed files.

desilinguist:
Looks fantastic! 

I just thought of something else though. From a grammatical point of view, `objectives` makes more sense since that matches how the other fields that take lists are named (`learners`, `featuresets`). So, perhaps, we want to allow all of the following:

- `objective = accuracy`
- `objectives = ["accuracy"]`
- `objectives = ["accuracy", "f1"]`

Essentially, we are letting people specify the old name (`objective`) but also allowing something that makes more sense grammatically (`objectives`). This is what we did in RSMTool so that people didn't have to go back and modify their old config files. t am sorry to keep adding things here. I am happy to make the changes myself in a separate PR if you'd prefer :)

bndgyawali:
I had implemented in the way you mentioned, using `objectives` and removing `objective` in my first pull request. I then reverted it all :) 

This needs the changes to be done on `config.py` and modifying a few `.cfg` files to test the changes. Please let me know if `experiment.py` looks good to you. Also, please mention how the preferences go if a same file contains `objectives` as well as `objective`. Or, throw an error if it contains both?

desilinguist:
Yeah, I know. I am sorry :/

However, I think I am suggesting something different than how you implemented things before. I think what you need to do is to have a little normalization function that basically replaces `objective` with `objectives` and the string with the list *before* you start parsing the config file. That is essentially what we do in RSMTool to be backwards compatible. And then the config parser only cares about `objectives` and lists. Does that make sense?

@aoifecahill, @dan-blanchard what do you think? I am also totally okay with just accepting this PR as is and moving my latest suggestion to a future issue.

aoifecahill:
But if we accept it as is, do we break backward compatibility? The future issue would still need to be resolved pretty quickly then?

desilinguist:
No, I think the way @bndgyawali has it now, backwards compatibility is preserved (since `objective=x` is still allowed). My suggestion was more to allow backwards compatibility AND to force people to the right thing in the future.

aoifecahill:
Ah, then I think that could maybe wait to a future issue?

desilinguist:
Yup, that works for me.

bndgyawali:
Sure, but, I think the issue @desilinguist  mentioned is simple to fix too. I will work on it on this branch.

desilinguist:
:+1:

desilinguist:
@bndgyawali do you think you will have time to work on this some more in the next week or so, given all the other things you have going on?

bndgyawali:
I should be able to send do this until the end of today. Will let you know soon.

desilinguist:
:+1:

bndgyawali:
This should now be in a good position for review.

desilinguist:
@dan-blanchard and @aoifecahill can you both please review this as well?

doc/run_experiment.rst
```diff
@@ -931,6 +933,14 @@ all share the following automatically generated prefix
     ``LEARNER``
         The learner the current results/model/etc. was generated using.
 
+    ``OBJECTIVE``
+        The objective function the current results/model/etc. was generated using.
+
+However, if ``objectives`` contains only one objective function, 
+the result, log, model, and prediction files will share the following prefix
```
dan-blanchard: Nitpick:  I'd drop the word "following" on this line.

doc/run_experiment.rst
```diff
@@ -920,7 +922,7 @@ Output files
 
 The result, log, model, and prediction files generated by run_experiment will
 all share the following automatically generated prefix
```
dan-blanchard: Might as well drop "following" here too. I think I had it originally before I added the section that broke down what each part meant on separate lines.

skll/config.py
```diff
@@ -222,6 +224,17 @@ def _setup_config_parser(config_path, validate=True):
...
+
+    # normalize objective to objectives
+    if config.has_option('Tuning', 'objective'):
+        objective = config.get('Tuning', 'objective')
+        objective = yaml.load(_fix_json(objective))
+        if isinstance(objective, string_types):
+            config.set('Tuning', 'objectives', "['{}']".format(objective))
+            config.remove_option('Tuning', 'objective')
+        else:
+            raise ValueError("objective should be a string")
```
dan-blanchard: Should be a `TypeError` here, not a `ValueError`.

skll/config.py
```diff
@@ -490,11 +503,16 @@ def _parse_config_file(config_path):
...
-    # what is the objective function for the grid search?
-    grid_objective = config.get("Tuning", "objective")
-    if grid_objective not in SCORERS:
-        raise ValueError('Invalid grid objective function: {}'
-                         .format(grid_objective))
+    # what are the objective functions for the grid search?
+    grid_objectives = config.get("Tuning", "objectives")
+    grid_objectives = yaml.load(_fix_json(grid_objectives))
+    if not isinstance(grid_objectives, list):
+        raise ValueError("objectives should be a "
```
dan-blanchard: `TypeError` here too.

skll/experiments.py
```diff
@@ -791,89 +792,95 @@ def run_configuration(config_file, local=False, overwrite=True, queue='all.q',
...
+            for grid_objective in grid_objectives:
+
+                # for the individual job name, we need to add the feature set name
+                # and the learner name
+                if len(grid_objectives) == 1:
+                    job_name_components = [experiment_name, featureset_name,
+                                           learner_name]
+                else: 
+                    job_name_components = [experiment_name, featureset_name,
+                                       learner_name, grid_objective]
```
dan-blanchard: Nitpick:  Please align `learner_name` with the opening bracket on this line (like is done three lines above).

dan-blanchard:
Thanks for putting this together! This looks great other than my few minor comments.

desilinguist:
Agreed, this looks good to me too. Thanks for accommodating all my comments, @bndgyawali  :)

bndgyawali:
:+1:

desilinguist:
One thing just occurred to me: if we want people to use the new syntax (`objectives`), then we should set an example and update the README as well as the examples?

dan-blanchard:
> One thing just occurred to me: if we want people to use the new syntax (objectives), then perhaps we should set an example and update the README as well as the examples (at least one of them)?

Good point.  All our examples should be updated.

bndgyawali:
I am updating it. I found that I need to modify them in `README.rst`, `docs/tutorial.rst` and `examples/*/*.cfg` files. Are there any more than this to be modified?

desilinguist:
@aoifecahill may be you can try the tutorial and one of the examples as part of your review to make sure they work?

skll/config.py
```diff
@@ -99,7 +99,7 @@ def __init__(self):
                                    'min_feature_count': 'Tuning',
                                    'models': 'Output',
                                    'num_cv_folds': 'Input',
-                                   'objective': 'Tuning',
```
aoifecahill: for backward compatibility, don't we want to leave this in?

bndgyawali: No, we do not leave it in. If the configuration file contains `objective`, objective will be converted into objectives and then it will start to validate the configuration file.

desilinguist: Actually, what will happen if someone puts an old style `objective=x` statement in the wrong section? Do we will raise the appropriate error or do we ignore that silently?

aoifecahill: I think that's why I was saying that it should stay in. For backward compatibility, we are still accepting that option, but we should also check that it is in the correct section?

bndgyawali: Having objective in `correct_section_mapping` will also need it in `defaults`, else `assert defaults.keys() == correct_section_mapping.keys()` will fail. 

Adding `objective` in `defaults` and `correct_section_mapping`, `super(SKLLConfigParser, self).__init__(defaults=defaults)` will add `objective` default value, and this will overwrite `objectives` from the logic I implemented. So, I could not find how I can do it. Suggestions please.

desilinguist:
Here's a possible solution:

You add both `objective` as well as `objectives` to both `defaults` as well as `correct_section_mapping`.  And then you implement the following when doing the normalization:

- If both `objective` and `objectives` are specified with default
values, we remove `objective` from the config.
- If `objectives` is specified with a non-default value but `objective`
has the default value, we remove `objective` from the config.
- If `objective` is specified with a non-default value but `objectives`
has the default value, we remove `objectives` from the config.
- If both `objectives` and `objective` are specified with non-default
values, we raise an error since user intent is not clear.

Would that handle all possible scenarios? @aoifecahill @bndgyawali ?

bndgyawali:
I think it handles the situations well. I will work on it.

desilinguist:
Great, thanks!

desilinguist:
@bndgyawali please make sure to include new tests for the other scenarios in that list.

tests/test_input.py
```diff
@@ -576,6 +588,83 @@ def test_config_parsing_bad_objective():
...
+    output_dir = join(_my_dir, 'output')
+
+    # make a simple config file that has a bad task
+    # but everything else is correct
+
+    config_template_path = join(_my_dir, 'configs',
+                                'test_objectives_error.template.cfg')
+    values_to_fill_dict = {'train_directory': train_dir,
+                           'log': output_dir,
+                           'results': output_dir}
```
desilinguist: It's a little strange to use a template pre-filled with the bad value here but not for `test_config_parsing_bad_objective_2`, no?

tests/test_input.py
```diff
@@ -576,6 +588,83 @@ def test_config_parsing_bad_objective():
...
+
+    config_template_path = join(_my_dir, 'configs',
+                                'test_objective_and_objectives_conflict_error.template.cfg')
+    values_to_fill_dict = {'train_directory': train_dir,
+                           'log': output_dir,
+                           'results': output_dir}
+    fill_in_config_options(config_template_path,
+                           values_to_fill_dict,
+                           'bad_objective_and_objectives')
+
```
desilinguist: Again, I am not sure why you are using a template pre-filled with bad values instead of using the regular config parsing template and passing in the values as part of `values_to_fill_dict`? And this way, you don't need the decorator right?

tests/test_input.py
```diff
@@ -860,7 +950,7 @@ def test_config_parsing_relative_input_paths():
                            'learners': "['LogisticRegression']",
                            'log': output_dir,
                            'results': output_dir,
-                           'objective': 'f1_score_macro'}
+                           'objective': 'f1_score_micro'}
```
desilinguist: why change this from `macro` to `micro`?

tests/test_input.py
bndgyawali: This is because, `test_config_parsing_bad_objective_2` is testing `objective` while this is testing `objectives`. And, template was used in this case because, at this moment `fill_in_config_options` does not have `objectives` in it.

desilinguist: I hate to say it but a better solution would be to add `objectives` to `fill_in_config_options` because that's something we would want anyway for future versions.

tests/test_input.py
bndgyawali: Because, `fill_in_config_option` calls `_setup_config_parser` and `check_config_parsing_key_error` also calls `_setup_config_parser`. In this case, the config file has `objective=accuracy`. So, in the first call to `_setup_config_parser`, it makes `objectives=['accuracy']`, and in the second pass, it has already filled up `objective=`f1_score_macro`, and this creates conflicts and throws error. So, I made the second objective to default so it passes successfully.

desilinguist: Fair enough. It doesn't matter what the objective is for this test anyway.

tests/test_input.py
bndgyawali: This is again because `objectives` not in `fill_in_config_options`.
I could instead have used `check_config_parsing_key_error` with all config sections/options already set in the config file instead of calling `fill_in_config_options`

desilinguist: Right, this should be easily fixed if you add `objectives` to `fill_in_config_options`.

tests/test_input.py
bndgyawali: Doing it.

tests/test_input.py
bndgyawali: using `fill_in_config_options` will not make us able to use decorator. Instead, using template config file will enable to use decorator. Please suggest which way to go through.

bndgyawali: Sorry, I was wrong. I am using decorator and deleted config file.

desilinguist:
okay, this now looks pretty close to perfect. I think some error messages and documentation sentences can be phrased a little better but I can take care of those and then merge.

Thanks again for working so hard on this! I really appreciate it.

desilinguist:
I ran this branch on our servers to make sure that it works with `gridmap` and it does :)

Issue #261, opened 2015-10-22
Add `pandas` data frame helper 


desilinguist:
It would be nice to have a helper method in the `FeatureSet` class (`FeatureSet.from_data_frame`, may be?) that would let you convert a pandas data frame to a FeatureSet. I already have the code to do this in another project so it shouldn't be a lot of work. I imagine it would really be useful to people who use the SKLL API but want to just use the data that they already have in a data frame.

Ideally, you would want all the main methods to accept data frames in addition to FeatureSets but that would be a lot of work. I think this is a nice workaround in the meanwhile.

desilinguist:
Sorry for all the mis-clicks for the milestones. GitHub apparently wants to keep a record of everything :)

dan-blanchard:
I think this is a nice idea, but it should be done in such a way that we don't actually need pandas as a prereq.

desilinguist:
Looking at my code, it will work just fine with the `DataFrame` object itself without relying on `pandas` to be available. However, I am not sure what the big deal is about using `pandas` as a prereq. It actually has lots of nice functionality.

dan-blanchard:
Eh, I just don't like when packages have requirements that are for optional things.  It's like how we'll currently use GridMap if it's installed, but it's not a requirement.

desilinguist:
I think at this point people who use `skll` but not `pandas` are going to be a very small minority. Plus, I can totally see adding a lot more functionality to `skll` by borrowing more from `pandas` built-in functionality.

dmnapolitano:
:+1:

dmnapolitano:
Question: do we want to make `pandas` a `skll` requirement?

dan-blanchard:
> I think this is a nice idea, but it should be done in such a way that we don't actually need pandas as a prereq.

:smile:

desilinguist:
Addressed by #292.

PR #292, opened 2016-02-12
Add a helper method to create FeatureSet from pandas data frames.

dmnapolitano:
Hello!  The changes in this pull request should resolve issue #261.  Thanks! :smile:


skll/data/featureset.py
```diff
@@ -359,3 +368,36 @@ def __getitem__(self, value):
...
+        :param name: The name of this feature set.
+        :type name: str
+        :param labels_column: The name of the column containing the labels (data to predict).
+        :type labels_column: str or None
+        :param vectorizer: Vectorizer that created feature matrix.
+        :type vectorizer: DictVectorizer or FeatureHasher
+        '''
+
+        if not _HAVE_PANDAS:
+            logger.warning(('pandas not installed.  Please install pandas or '
```
dan-blanchard: This should just raise a `ValueError` if they don't have pandas, because that's the only time anyone should use `from_data_frame`.

dan-blanchard: Actually, I take that back.  You're not actually using `pandas` anywhere in here, so why even bother checking if you can import it?  Duck typing should prevail here.  (How is there no duck emoji?)

dmnapolitano: Ok word, we had the same thoughts here.  To be revised :smile:   (Also yeah the lack of duck emoji makes no sense :shipit: )

tests/test_featureset.py
```diff
@@ -845,3 +846,68 @@ def test_convert_featureset():
...
+    if with_labels:
+        df['y'] = y
+        current = FeatureSet.from_data_frame(df, featureset_name, labels_column='y')
+    else:
+        current = FeatureSet.from_data_frame(df, featureset_name)
+
+    return (expected, current)
+
+
+@attr('have_pandas')
```
dan-blanchard: Huh, didn't know about that feature.

skll/data/featureset.py
```diff
@@ -18,6 +19,7 @@
 
 from skll.data.dict_vectorizer import DictVectorizer as NewDictVectorizer
 
+logger = logging.getLogger(__name__)
```
dan-blanchard: You can drop this logging stuff because we're not issuing a warning anymore.

desilinguist:
I really like this solution but do really need to run *all* the tests in the `PANDAS=true` environments? Seems like its presence 's won't affect anything except the specific test that we want to run. Am I wrong?

skll/data/featureset.py
dmnapolitano: Ah right, thanks for the reminder :smile: :+1:

dmnapolitano:
@desilinguist Hmmmmmmmm...so in Travis, when `WITH_PANDAS` is `True`, only run the `pandas` tests and skip the others?  I can make that happen :smile: :panda_face:

skll/data/featureset.py
```diff
@@ -359,3 +359,31 @@ def __getitem__(self, value):
...
+        if labels_column:
+            labels = list(df[labels_column])
+            features = [{col : row[col] for col in row.keys() if col != labels_column}
+                        for (i, row) in df.iterrows()]
+        else:
+            labels = None
+            features = [dict(row) for (i, row) in df.iterrows()]
+
+        return FeatureSet(name, list(df.index), labels=labels,
+                          features=features, vectorizer=vectorizer)
```
desilinguist: I believe, there's a more efficient way to do this like in RSMTool:

```python
if labels_column:
    feature_columns = [column for column in df.columns if column != labels_column]
    labels = df[labels_column].tolist()
else:
    feature_columns = df.columns
    labels = None

features = df[feature_columns].to_dict(orient='records')
return FeatureSet('train', ids=df.index.tolist(), labels=labels, features=features, vectorizer=vectorizer)
```

desilinguist:
Since this is a major new feature, it would be nice to:

1. Update the documentation to explicitly mention it in an appropriate place. 
2. Provide an example showing the feature in action.

dmnapolitano:
Ok, I will try to finish this up today and Sunday. :smile:

desilinguist:
Thanks! I think the code part is good (assuming Travis agrees :) ) . Now we need a nice example (perhaps a panda-ized version of the Titanic example?) and updated documentation.

desilinguist:
All this needs is an example and updated documentation. We will include this in the next release.

desilinguist:
@dmnapolitano do you have time to work on this for the 1.2.1 release which we want to do in a week or so? I believe all it needs is an example and updated documentation and making sure that the tests still pass.

dmnapolitano:
Hmmm yeah, I believe you're right. 👍   Ok, let me see what I can do by Monday.  If the answer is "nothing" I'll be sure to let you know. 😬   Thanks!

dmnapolitano:
Ok, I think this is all set, but let me know what you think of course 😄 👍

desilinguist:
@dan-blanchard can you take a look as well please? I will too.

tests/test_featureset.py
```diff
@@ -845,3 +846,68 @@ def test_convert_featureset():
                                                           '.csv', '.arff',
                                                           '.libsvm'], 2):
         yield check_convert_featureset, from_suffix, to_suffix
+
+
+def featureset_creation_from_dataframe_helper(with_labels):
+    """
+    Helper function for the two unit tests for FeatureSet.from_data_frame().
+    Since labels are optional, run two tests, one with, one without.
```
aoifecahill: ``vectorizer`` is also optional. It might be nice to add tests for with/without vectorizer as well?

dmnapolitano: Ok 👍🏼  I don't know much about vectorizers, honestly.  Aren't there quite a few of them one could use?

aoifecahill: Take a look at how it is created in the skll ``utils.py`` method ``make_classification_data`` and see whether you could apply that here too on your data?

aoifecahill:
This looks ok to me. Any other comments?

desilinguist:
Looks good to me too. @dan-blanchard do you have anything?

doc/api/quickstart.rst
```diff
@@ -3,15 +3,22 @@ Quickstart
...
     example_reader = Reader.for_path('myexamples.megam')
     train_examples = example_reader.read()
 
+    
+Or, work with an existing ``pandas`` ``DataFrame``::
 
-Train a linear svm (assuming we have `train_examples`)::
+    from skll import FeatureSet
+
+    train_examples = FeatureSet.from_data_frame(my_data_frame, 'A Name for My Data', data_labels)
```
aoifecahill: This example is a bit misleading. When I read it, I assumed `data_labels` was a list of labels, not the name of a column in the `my_data_frame` df.

dmnapolitano: Right, sorry!  I forgot how my own code works 😞   I'll update this now, thanks.

skll/data/featureset.py
```diff
@@ -363,3 +363,33 @@ def __getitem__(self, value):
...
+        Will raise an Exception if pandas is not installed in your environment.
+        `FeatureSet` `ids` will be the index on `df`.
+
+        :param df: The pandas.DataFrame object you'd like to use as a feature set.
+        :type df: pandas.DataFrame
+        :param name: The name of this feature set.
+        :type name: str
+        :param labels_column: The name of the column containing the labels (data to predict).
+        :type labels_column: str or None
+        :param vectorizer: Vectorizer that created feature matrix.
```
desilinguist: This description of `vectorizer` seems weird? Shouldn't this be the vectorizer you *want* to use when creating the featureset?

dmnapolitano: Honestly I'm not really sure.  I copied it from line 44. 😬

desilinguist: @aoifecahill this will be the vectorizer used to create the featureset and NOT one that has already been used for something, right?

aoifecahill: right, it's the one that will be used to create this new featureset. It could have been used for something else though I think (e.g. if you want to apply a pre-existing vectorizer to a new test set)

desilinguist: ah, right. In any case, @dmnapolitano can you fix the docstring?

dmnapolitano: Ok, should I change line 44 too?

aoifecahill: No, line 44 is correct.

dmnapolitano: Ok, then can you help me understand why they're different?  In both cases we're creating a `FeatureSet`, so why would the `vectorizer` do something different in this case?

desilinguist: I think both can docstrings can just say “Vectorizer to use for generating feature matrix”?

dan-blanchard:
Looks good to me.

dmnapolitano:
Thanks, everyone! 🎉

PR #294, opened 2016-02-12
added note about FeatureSet equality

aoifecahill:
Minor update to FeatureSet class documentation. Not sure whether it should go anywhere else.

### DIFF
--- a/skll/data/featureset.py
+++ b/skll/data/featureset.py
@@ -24,8 +24,8 @@ class FeatureSet(object):
     """
     Encapsulation of all of the features, values, and metadata about a given
     set of data.
-    WARNING: FeatureSets can only be equal if the order of
-    the instances is identical because these are stored as lists/arrays.
+    warning:: FeatureSets can only be equal if the order of the instances is
+    identical because these are stored as lists/arrays.
 
     This replaces ``ExamplesTuple`` from older versions.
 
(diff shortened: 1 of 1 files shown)


skll/data/featureset.py
```diff
@@ -23,7 +23,8 @@ class FeatureSet(object):
 
     """
     Encapsulation of all of the features, values, and metadata about a given
-    set of data.
+    set of data. Note that FeatureSets can only be equal if the order of
```
dan-blanchard: Nitpick:  Should probably use sphinx syntax for this:

```py
.. note::
    FeatureSets can only be equal if the order of the instances is identical 
    because these are stored as lists/arrays.
```

Might even be worth doing `.. warning`.

skll/data/featureset.py
```diff
@@ -24,6 +24,8 @@ class FeatureSet(object):
     """
     Encapsulation of all of the features, values, and metadata about a given
     set of data.
+    warning:: FeatureSets can only be equal if the order of the instances is
```
dan-blanchard: Almost!  You want this to be:

```rst
Encapsulation of all of the features, values, and metadata about a given
set of data.

.. warning:: 
    FeatureSets can only be equal if the order of the instances is identical 
    because these are stored as lists/arrays.
```

Issue #285, opened 2016-02-12
Non-integer labels for classification


aloukina:
Non-integer numeric labels for classification experiments are treated as integers without any warning. For example, [1, 1.6, 1] becomes [1, 1, 1].

aloukina:
This is not true.

aoifecahill:
Phew :)

aloukina:
It does happen if you generate test_fs and pass it to learner.evaluate() directly: not sure if this scenario is possible.

desilinguist:
That seems like something you would do a lot when using the API. We do it in RSMTool (albeit for regression)

desilinguist:
@aoifecahill confirmed that this is not a problem and just to be sure we added another test for this in #295.

aloukina:
For future reference: the float to integer conversion happens if one creates data using make_classification_data and then changes the labels by using indices:
`train_fs, test_fs = make_classification_data(num_labels=3, train_test_ratio=0.8)
test_fs.labels[-1]=1.5
`

PR #295, opened 2016-02-12
add test for floating point labels in classification

aoifecahill:
This is to convince ourselves that the situation in #285 does not happen.


desilinguist:
This looks good to me

dan-blanchard:
:+1:

desilinguist:
Ah, you probably want to update the `tearDown()` method to clean up the files that you produce in this test.

PR #290, opened 2016-02-12
Correct treatment of the unseen labels in the training set

aloukina:

PR #296, opened 2016-02-13
[WIP] Modularize pretty printing of confusion matrix

bwriordan:


desilinguist:
I think there is a dependency between this PR and #290. So, we need to merge one of them first and then then update the other one.

aloukina:
There should also be a test to ensure that the labels are printed correctly when there are unseen labels in the test set. Currently SKLL refers to classification classes by either actual label or by the indices assigned to sorted labels. When we add new unseen labels the ordered indices no longer correspond to the sorted labels so there is a potential for a bug here.

desilinguist:
I am going to close this for now.

PR #297, opened 2016-02-13
Add warnings for conflicts between fixed params and param grids

mulhod:
Added warnings for conflicts between fixed parameters and parameter grids and for when grid search is set to False and paramater grids are specified.


skll/config.py
```diff
@@ -474,6 +474,26 @@ def _parse_config_file(config_path):
...
+            if overlap_params:
+                logger.warning('Both "fixed_parameters" and "param_grids" were'
+                               'specified or the default values were used for '
+                               '{}, which resulted in conflicts between '
+                               'certain parameter(s): {}. Values passed in via'
+                               '"param_grids" will take precedence in these '
+                               'cases.'.format(learner,
+                                               ', '.join(overlap_params)))
+    else:
+        if param_grid_list or fixed_parameter_list:
```
desilinguist: What's the problem with specifying `fixed_parameter_list` even if we aren't doing grid search?

skll/config.py
```diff
@@ -474,6 +474,26 @@ def _parse_config_file(config_path):
...
+    # is True
+    if do_grid_search:
+        for learner, fixed_params, params in zip(fixed_parameter_list,
+                                                 param_grid_list,
+                                                 learners):
+            overlap_params = set(fixed_params).intersection(set(params))
+            if overlap_params:
+                logger.warning('Both "fixed_parameters" and "param_grids" were'
+                               'specified or the default values were used for '
+                               '{}, which resulted in conflicts between '
```
desilinguist: What's an example of the second condition ("default values were used ..."?

skll/config.py
mulhod: Fixed it

skll/config.py
mulhod: As I understand it, specifying `grid_search` and then not specifying `param_grids` will default to the default parameter grids. So, let's say that someone wants to do grid search and doesn't specify the parameter grids. Then, they go to specify their fixed parameters and they don't really know what conflicts there may be between the defaults and the values they pick (probably rare, but still could happen). There could be a conflict between a parameter value they specified as fixed and a paramter grid value they did not specify at all. Does that make sense?

aoifecahill: You mean the situation where somebody specifies some fixed parameters but leaves on grid search (without specifying the grid)? I think the warning message could be more general, something like: "There is a conflict between the fixed parameters specified and the parameters to be searched during grid search. The fixed parameters will be used."?

mulhod: It seems like the "param_grids" values actually take precedence, not the "fixed_parameters". Maybe my warning message isn't very clear, but it seems like it's saying exactly what you're saying above. I'm not understanding how that's more general. In any case, I will rephrase it.

mulhod: Ok, I think I get the general comment now. Was interpreting it as "cover more situations than you're covering".

desilinguist:
Okay, @aoifecahill and I talked about this a bit and we think this is a little bit tricker:

1. If users specify a parameter in `fixed_params` but also specified the *same* parameter as part of their parameter grid to search, that should actually be an error since it's not clear what their intent is. 

2. However, if users specify a parameter in `fixed_params` but have *only* specified `grid_search=True` and the default parameter grid happens to include said parameter, there should be a warning saying that the fixed parameter will be ignored. But that warning should also tell the user how to avoid this if they want - by specifying a parameter grid explicitly which does not include this parameter.

3. We should also update the documentation for `fixed_params` to indicate all of the above.

Thoughts?

mulhod:
Seems pretty clear to me. I'll update this shortly.

dan-blanchard:
> However, if users specify a parameter in fixed_params but have only specified grid_search=True and the default parameter grid happens to include said parameter, there should be a warning saying that the fixed parameter will be ignored. But that warning should also tell the user how to avoid this if they want - by specifying a parameter grid explicitly which does not include this parameter.

To me it would make the most sense in this case for the fixed parameter to win out over the default parameter grid.  We could just modify the grid for them automatically and likely get them what they wanted.

desilinguist:
Hmm, yeah I agree. That's a better way of doing git.

dan-blanchard:
Yeah, I agree about the warning.

mulhod:
Couple observations:

- First, I noticed an error in the check the way I implemented it and it's been fixed: I was zipping some things together and got the sequence of the zipping wrong. So, I just wanted to mention this in case it's been noticed.
- More importantly, I have realized that the original way I was doing this wasn't correct. I was treating it as if `param_grid_list` would be populated with the default values before the point when it would get to the check at the center of this PR. However, the default values only get substituted in later on if `param_grid_list` either just wasn't specified at all or if some of the parameter grids in it are equal to an empty list, in which case it is my understanding that the default parameter grid is used. This was written this way in order to make it so that users could specify some, but not all parameter grids. Having an empty list was the same as saying, "Use the default grid". Given that the value of `param_grid_list` will either be None (if not specified) or a list of parameter grid dictionary lists (which might contain some empty lists), my check here will have to be a lot more involved -- _if `fixed_parameters` and `grid_search` are specified_. My check will basically need to look at the default values and see if there are conflicts and, in some situations, actually override values. And, since the would have happened after this point the way things were, it will now happen here.

mulhod:
Ok, I've made a bunch of changes and I think this is finished -- or, at least, I don't want to work on this further until somebody else takes a look because I had to change quite a few things and I want to know first if this is what you had in mind. As I mentioned earlier, in order to intelligently resolve conflicts, logging warnings sometimes, raising exceptions others, or doing some things silently, the approach had to be much more involved than it was initially. Partially this is due to the scope of the PR changing a little after further discussion. Anyway, tests pass locally, so I'm fairly certain that it works as is.

doc/run_experiment.rst
```diff
@@ -507,7 +507,9 @@ fixed_parameters *(Optional)*
 
 List of dicts containing parameters you want to have fixed for each
 classifier in :ref:`learners` list. Any empty ones will be ignored
-(and the defaults will be used).
+(and the defaults will be used). If there is a conflict between parameters
+specified in :ref:`param_grids` and :ref:`fixed_parameters`, :ref:`param_grids`
+values will take precedence.
```
desilinguist: Isn't this the opposite of what we finally agreed on (after @dan-blanchard's [comment](https://github.com/EducationalTestingService/skll/pull/297#issuecomment-185818068)?

mulhod: I haven't dealt with updating the documentation yet after making the most recent changes.

desilinguist: Ah, okay. I just wanted to make sure that the code is doing the right thing and it's just the documentation that's out of sync.

skll/config.py
```diff
@@ -479,6 +495,102 @@ def _parse_config_file(config_path):
...
+        # `fixed_parameter_list` and simply remove the conflicting values from
+        # the default parameter grids and use the modified defaults, but log a
+        # warning
+        else:
+            param_grid_list = \
+                [_find_default_param_grid(_LEARNER_NAMES_TO_CLASSES[learner_name])
+                 for learner_name in learner_names]
+            for i, (learner,
+                    fixed_params,
+                    params_grids) in enumerate(zip(learners,
```
mulhod: Just realized I don't actually need `learners` here, so that will be removed.

mulhod:
So, it seems that some of my tests are failing, but not the ones that are really for the core of this PR. Specifically, they are the tests for the learner names in the config file. I thought that I could simply tests all the input names and then raise an exception if they weren't in the list of learner classes. But, I wasn't accounting for custom learners. What I'm wondering is: did you not have this type of check specifically to allow custom learners? I could change my exception to a warning that simply says that the user is using an unrecognized/custom learner. Or I could take it out. Also, the reason I had added that in there was because I needed to be able to refer to a learner's corresponding class (without using `eval`) to get its default parameter grid. I could change that check below to look for the parameter grid and if none is found (i.e., because it's a custom learner), then obviously don't do the usual check since there will be no parameter grid.

desilinguist:
Yeah, I think we weren't checking the learners's identity because people could write custom learners. I think this PR requires a bit of careful reviewing since the scope has expanded. I would like all of us to think about this carefully and reach a consensus. So, this can probably go into either 1.2.1 or 1.3, whatever release we do next.

mulhod:
The check for a valid learner type, which my changes fail on due to the fact that users can provide custom learners, could actually be resolved just simply by checking if the learner ends with ".py", couldn't it? I think the error that I have put in could be kept without much modification.

mulhod:
Not that it matters much, but I don't quite understand how the coverage check failed if all I did was merge in the master branch. I know the coverage could have gone down because it was just lower in the master branch or due to updates here, but there haven't been any since the last actual (non-merging in master) commit. It seems like it shouldn't have failed, though.

aoifecahill:
Could it have been that you didn't run with the fixed coveralls until today? (that was one of the recent updates to master). Essentially the coveralls checks were not being done for a while because the package was out of date.

mulhod:
Ohh, ok, maybe that's it @aoifecahill

mulhod:
Is there anything that this is still waiting on? Anything else that I can do on my end?

desilinguist:
The coverage decrease is not good. May be, merge in `master` after we merge in a couple of other PRs so that the tests run again?

mulhod:
It seems to me that the only uncovered portion could be easily covered simply by adding a unit test. That would probably actually increase coverage. But is that the only issue?

desilinguist:
It looks like you *are* adding a unit test though, right?

mulhod:
I've already done that, yes. I just mean that another one to deal with the final portion that is not covered looks like it's the only thing that remains.

desilinguist:
Feel free to do that.

mulhod:
What I want to know is: is that the only issue? I don't want to spend time on something that's not going to be merged in for some other reasons.

mulhod:
If there are no comments, I think this is ready. There was an issue with the coverage but that's been resolved.

mulhod:
I forgot that there is actually some documentation that needs to be updated slightly (see top of thread above). I was waiting for there to be some discussion about the changes before I did that. But I'll just go ahead and change it now.

doc/run_experiment.rst
```diff
@@ -506,8 +506,18 @@ fixed_parameters *(Optional)*
 """""""""""""""""""""""""""""
 
 List of dicts containing parameters you want to have fixed for each
-classifier in :ref:`learners` list. Any empty ones will be ignored
-(and the defaults will be used).
+learner in :ref:`learners` list. Any empty ones will be ignored
+(and the defaults will be used). If there is a conflict between parameters
+specified in :ref:`param_grids` and :ref:`fixed_parameters`, an exception
+will be raised (unless the conflicting values are the same). Furthermore,
```
aoifecahill: they're not really conflicting if they're the same :)

doc/run_experiment.rst
```diff
@@ -506,8 +506,18 @@ fixed_parameters *(Optional)*
...
+since default parameter grids can be used either by not specifying
+parameter grids at all or by specifying blank lists for specific
+parameter grids corresponding to different learners (i.e., in cases where
+you want to use default parameter grids for some learners, but for others
+you want to specify the parameter grids yourself), there is a potential for
+conflicts between fixed parameters and default parameter grid values that
+are not obvious initially. When such conflicts arise, the fixed parameter
+values will take precedence and a warning will be logged.
```
aoifecahill: These feel like very similar situations (there is a conflict between `fixed_parameters` and `param_grids`. In one case an exception is raised, but in the other a warning is issued. This feels a little strange though I see how the cases are subtly different.

skll/config.py
```diff
@@ -338,6 +353,17 @@ def _parse_config_file(config_path):
     custom_learner_path = _locate_file(config.get("Input", "custom_learner_path"),
                                        config_dir)
 
+    # Check that the learners are valid learner types (or that they are custom
+    # learners, i.e., ending in ".py")
```
aoifecahill: I think this comment about engine in ".py" is out of date now?

desilinguist:
I really would like @dan-blanchard to chime in on this PR. Dan, whenever you get a chance.

doc/run_experiment.rst
mulhod: I agree, maybe "conflict" is the wrong word. I just meant the situation where a parameter is both referred to in the fixed parameters and in the parameter grid. This is a bit of an unusual situation, probably. And the only way they can agree is by specifying a list containing a single value in the parameter grid. To me, that seems like the same thing as just fixing its value. Anyway, maybe I can just rephrase it?

mulhod: I agree, "conflict" might not be the right word here. I just meant that the same parameter is specified in both the fixed parameters and in the parameter grids. If it's just a single value inside a list in the parameter grid and the value is the same as the fixed parameter, it doesn't matter, right? It seems to me that that is basically another way to fix a parameter's value. Anyway, the main point is that I wanted to recognize that situation and not raise an exception there. I can think about rephrasing it.

skll/config.py
dan-blanchard: This check seems a bit overly complex when you consider that we already have [these checks](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L199-L205) for custom learners and [this](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L546) will raise a `KeyError` if a learner is not valid.

Is the thought process that we want to raise an exception here before we get there?  If so, it seems like there's got to be a simpler way to do this other than needing this whole `_LEARNER_NAMES_TO_CLASSES` thing.  You could probably even just look in just do:

```py
unrecognized_learner_names = {learner_name for learner_name in learner_names 
                              if 'skll.learner.{}'.format(learner_name) not in globals()}
```

skll/config.py
```diff
@@ -33,6 +35,19 @@
                                             'none'])
 
 
+def _transform_learner_name(learner_name):
```
dan-blanchard: I think you should just snatch the class out of `globals()` rather than making your own dict and using this transformer thing.

tests/test_input.py
```diff
@@ -1132,3 +1138,253 @@ def test_setting_fixed_parameters():
...
+     min_feature_count, grid_search_jobs, grid_search_folds, cv_folds, save_cv_folds,
+     do_stratified_folds, fixed_parameter_list, param_grid_list, featureset_names,
+     learners, prediction_dir, log_path, train_path, test_path, ids_to_floats,                          
+     class_map, custom_learner_path) = _parse_config_file(config_path)
+
+    # The fixed parameter value that was specified should remain as is
+    # while the conflicting default parameter value will be removed from
+    # the parameter grid
+    assert fixed_parameter_list[0]['alpha'] == 0.01
+    assert param_grid_list[0][0].get('alpha') == None
```
dan-blanchard: `== None` should be `is None` here.

tests/test_input.py
```diff
@@ -1132,3 +1138,253 @@ def test_setting_fixed_parameters():
...
```
dan-blanchard: `== None` should be `is None` here.

dan-blanchard:
I'm cool with the general idea of this PR, and it seems tested enough that I'll trust the occasionally  complex logic—which is totally necessary because of the all combinations you need to check—but there are a couple things that could be simplified that I outlined in my comments.  

I also agree that it should be a warning when you override the default parameter grid or a blank one, but not when there's an actual user-specified conflict.

skll/config.py
mulhod: I have removed the check. I think I was just adding in a check because I didn't know it was done elsewhere. It might be nice to have a check here so that mistakes get recognized immediately, but I don't have a strong opinion about this.

However, before I removed it, I just want to mention that I did try the other method you provided an example of to get the unrecognized learner types. I couldn't get it to work correctly. `globals()` does not seem to include anything of the form `skll.learner.x` where `x` is the learner name. Nor does it include just the learner name itself. I tried making modifications to it, but I could not get this to work. I mention this here mainly because it will tie into what I have to say later about using `globals()` to get the classes that correspond to learner names.

skll/config.py
mulhod: As I mention in a separate comment, I liked the idea of simply using `globals()` to do this and it would have made things a lot simpler and cleaner. However, it seems to me that `globals()` does not contain any entries that correspond to learner names. I can get access to learner names via `dir(skll.learner)`, so there is a way that I can confirm that a learner name is simply defined in `skll.learner`. However, to get the corresponding class, there's no obvious way (to me, at least) to do that without something like `eval('skll.learner.{}'.format(learner_name))`. While this probably isn't that dangerous (despite the use of `eval`) since it would be *after* confirming that the `learner_name` part is at least defined in `skll.learner`, I don't think it looks good or is any better than the way I did it by creating a new `_LEARNER_NAMES_TO_CLASSES` dict, etc.

Please correct me if I'm wrong.

mulhod:
It's been awhile, but I took the opportunity to merge in recent changes and see if this still passes.

Any interest in actually dealing with this? If not, it seems like maybe the issue should just be closed or something and this PR declined as "Will not do".

desilinguist:
I am still not particularly enthused by the idea of adding such complicated logic to the code base. Given that no one has really run into this issue in the wild, perhaps we can simplify the PR to just do the following:

1. Show the warning if `param_grids` are specified but `grid_search` is false (which is already in this PR).
2. Show a warning saying that `fixed_parameters` will take precedence over `param_grids`, if both are specified. And let the user figure out the conflicts and fix the config file.

I know you did a lot of work for this PR @mulhod and I apologize for making you do said work but that's how I feel at this point. Thoughts?

mulhod:
The coverage decreased, but all that was added (now that some things have been removed) are basically logging statements. Tests could be added to check that the statements are logged, but I'm wondering if it's necessary or more trouble than it's worth in this case.

aoifecahill:
wouldn't we want to check that the correct warnings are being logged for the various scenarios?

mulhod:
It would be good, sure, but there are no checks for any of the other logging statements in similar situations. Also, it seems to me that it would be better if all of the warnings were actual warnings (that could be written to the log as well). That would be easier to test.

aoifecahill:
Ah yes, true. But since PR #380 has been merged, shouldn't that now be possible? i.e. using proper logging and verifying the logging messages?

mulhod:
Yeah, I see the case where it was used to verify the number of folds, etc. I think it's possible. I just wonder about 1) how worthwhile it is and 2) whether the warnings should be more like actual warnings. I can get around to adding some tests using the log statements for now.

aoifecahill:
Yes, I agree that the warning should be more like actual warning following the logging changes that @desilinguist implemented recently.

mulhod:
I have now added some tests that search for the warning in the log file. What I mean about using actual warnings, though, was about using the `warnings` module directly. I guess it doesn't matter since it's not extremely different in the end.

I think this is ready to go again.

skll/config.py
```diff
@@ -582,6 +581,23 @@ def _parse_config_file(config_path, log_level=logging.INFO):
     # using the defaults?
     do_grid_search = config.getboolean("Tuning", "grid_search")
 
+    # Check if `param_grids` is specified, but `grid_search` is False
+    if param_grid_list and not do_grid_search:
+        logger.warning('"param_grids" was specified despite the fact that '
```
desilinguist: I'd rephrase this to state simply that `Since "grid_search" is set to False, the specified "param_grids" will be ignored.`.

skll/config.py
```diff
@@ -582,6 +581,23 @@ def _parse_config_file(config_path, log_level=logging.INFO):
...
+                       '"grid_search" was not specified or specified as '
+                       'False.')
+
+    # Warn user about potential conflicts between parameter values
+    # specified in `fixed_parameter_list` and values specified in
+    # `param_grid_list` (or values passed in by default) if
+    # `do_grid_search` is True
+    if do_grid_search and fixed_parameter_list:
+        logger.warning('"grid_search" is set to True and "fixed_parameters"'
```
desilinguist: Minor rephrasing: `Note that "grid_search" is set to True and "fixed_parameters" is also specified. If there is a conflict between the grid search parameter space and the fixed parameter values, the fixed parameter values will take precedence.`

PR #303, opened 2016-02-19
[WIP] Restore gridmap testing to travis builds

desilinguist:
- I tried to mirror the same changes that @dan-blanchard pushed in his [latest PR](https://github.com/pygridtools/gridmap/pull/69) to `gridmap`. 
- I also used the `@attr` trick discovered by @dmnapolitano to only run the 5 tests that will actually run on the grid when `GRIDMAP=true`.
- However, this doesn't seem to work and still keeps failing with the same `DeniedByDrmException` like before.


desilinguist:
@dan-blanchard, any thoughts on why the same thing works for `gridmap` but not for `skll`?

dan-blanchard:
Nothing is jumping out at me, since it looks like you restored things to how they were before and added my little changes for GridMap.  I can look through it in more detail next week.

desilinguist:
Okay, thanks!

desilinguist:
Not sure this will ever happen so closing it for now.

PR #306, opened 2016-02-23
`fit()` in `rescaled()` decorator should return `self`

desilinguist:
Currently the `fit()` method in the `rescaled()` decorator does not return `self` which is generally what `fit()` methods in `scikit-learn` should do. 

I fixed that and also modified the `test_rescaling()` regression test to also test rescaled learners without using grid search. So, we have new tests to check for this functionality which we didn't have before which is probably why this bug slipped through,

### DIFF
--- a/skll/learner.py
+++ b/skll/learner.py
@@ -321,6 +321,8 @@ def rescaled(cls):
             self.y_mean = np.mean(y)
             self.y_sd = np.std(y)
 
+        return self
+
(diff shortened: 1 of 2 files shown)


desilinguist:
@aoifecahill can you also look at this please?

aoifecahill:
Looks good to me, thanks!

dan-blanchard:
Looks good to me too!

PR #308, opened 2016-02-23
Upgrade scikit-learn to 0.17.1

desilinguist:
While we were working on version 1.2 of SKLL, the `scikit-learn` folks released a [minor upgrade](http://scikit-learn.org/stable/whats_new.html) from 0.17.0 to 0.17.1. It most likely will not impact any of our tests but has some important bug fixes so we want to include it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
   - conda config --add channels desilinguist
   - conda update --yes conda
 install:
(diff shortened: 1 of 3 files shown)


desilinguist:
I ran this branch on my Mac as well as on one of our servers with `gridmap`. They both passed so I am going to merge this.

Issue #282, opened 2016-02-12
New update to conda numpy/scipy packages breaks travis builds


desilinguist:
Recently, conda switched to using (the so-far proprietary) `mkl` backend as the default for their numpy and scipy packages. However, travis builds break when using this backend (see [here](https://travis-ci.org/EducationalTestingService/skll/jobs/108818038)). The fix is to force numpy and scipy to use the `openblas` backend by installing the `nomkl` metapackage. I am working on a hotfix now.

desilinguist:
@dan-blanchard please let me know if I am misunderstanding what we need to do to fix it (if you have the time and inclination to investigate :) )

desilinguist:
Addressed by #284.

PR #309, opened 2016-02-23
Add new conda-recipes for unix and windows

desilinguist:
The current conda recipe will not work. See #282. 

So, we need to have two separate recipes - one for Unix that does not use MKL and one for Windows where we have to use MKL.

### DIFF
--- a/conda-recipe/unix/skll/meta.yaml
+++ b/conda-recipe/unix/skll/meta.yaml
@@ -1,0 +1,92 @@
+package:
+  name: skll
+  version: 1.1.1
+
+source:
+  path: ../../../../skll
+
+build:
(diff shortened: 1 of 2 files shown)


dan-blanchard:
Sorry I didn't get to comment on this until now, but we should really just stick to one recipe.  You can accomplish the exact same thing with [preprocessing selector comments](http://conda.pydata.org/docs/building/meta-yaml.html#preprocessing-selectors).

Add `# [not win]` to the ends of the lines related to `nomkl` and you'll be fine.  No need to have two separate recipes.

desilinguist:
I tried that it didn't work. I will try again.

PR #312, opened 2016-03-01
add auto to gamma tuning grid for SVR/SVC

aoifecahill:
Add "auto" to the SVR and SVC gamma grids to try and improve performance of the grid search.


dan-blanchard:
:+1:

PR #319, opened 2016-05-13
Replace PyYAML with a new library and allow scientific notation for floating point numbers

desilinguist:


desilinguist:
I am a little confused about the really old commits showing up in this PR. I will create a different branch and reopen a new PR.

PR #320, opened 2016-05-13
Replace PyYAML with a new library to allow floating point numbers in scientific notation.

desilinguist:


dan-blanchard:
What's the `conda_requirements` file for? Don't you just want to update the recipe directly?

desilinguist:
I am about the update the recipe. But one of the things all of us are used to doing when setting up a SKLL development environment is `conda create -n sklldev --file requirements.txt`. Given the discrepancy in the `ruamel_yaml` package name between PyPI and conda, we can't do that anymore. Hence `conda_requirements.txt`.

dan-blanchard:
If they ever merge my `conda-env` PR (conda/conda-env#172), you won't need to maintain two separate files anymore.

PR #321, opened 2016-05-19
Fix conda recipes

desilinguist:
- This branch was started to see if we can merge the unix and windows condo-recipes into a single recipe. However, it looks like `conda` still doesn't properly support the `[unix]` and `[not win]` selectors and I am not confident enough to switch to `mkl` for all platforms. So, this will have to wait.
- However, while I was doing this, I realized that even building the split conda recipes will fail since `conda build` runs `python setup.py install` which will then try to install `ruamel.yaml` from `requirements.txt` since conda installs `ruamel_yaml` instead. So, I basically, built our own conda packages (for all platforms) called `ruamel.yaml` and updated the conda recipes to use that instead.


dan-blanchard:
I don't see why this package name change is necessary.  Package names are different for conda and pip/setuptools all the time.  Once things are installed, the actual package name should match the pip/setuptools one, and pip will just notice that it's already installed.

desilinguist:
Yeah, that would be ideal but as I said, `conda build` is unable to figure out that the `ruamel_yaml` package it has already installed from `meta.yaml` is the same package as `ruamel.yaml` from `requirements.txt` and so it tries to install that and then fails.

dan-blanchard:
Apparently they screwed that recipe up, which is impressive.

Anyway, it's available with the proper name on the `conda-forge` channel.  https://github.com/conda-forge/ruamel.yaml-feedstock.  I would recommend adding that channel to your channel list.  They're doing lots of good things with that project.

desilinguist:
Ah, I had heard about conda-forge but never looked into in much detail. It does look pretty good.

desilinguist:
The coverage decrease is obviously spurious so I will merge this unless you have any objections @dan-blanchard ?

desilinguist:
Thanks, @dan-blanchard!

PR #323, opened 2016-05-19
updating for logistic regression with no intercept

aoifecahill:
Adds in a check for if there is no intercept (it shows up as a bare float of 0.0, compared to a list of floats if an intercept was used). 

Also adds a test.


desilinguist:
This looks good to me.

Issue #231, opened 2015-03-03
When scikit-learn 0.18 comes out, we need to update our F1 metrics in __init__.py


dan-blanchard:
scikit-learn/scikit-learn#4192 makes it so that setting `pos_label` to `None` is no longer necessary with binary datasets (because you'll have to use "binary" as the value for `average` to get the one-class version it returns now).  This will become the default behavior in 0.18, so we'll need to update things then.  This was brought up all the way back in #3.

desilinguist:
Addressed by #330.

PR #330, opened 2016-12-09
Upgrade scikit-learn to v0.18.1

desilinguist:
- The major change in this version is that most the cross-validation and grid-search stuff has been moved to `sklearn.model_selection` AND the interface of the cross-validation generators has changed. They are no longer utterable directly but rather you have to initialize and then call the `.split()` method on them when you want the actual train/test indices.

- There are also some other major updates in the underlying algorithms themselves and in `make_classification()` which required changing the expected values in several of the tests. 

- I also address #231 by updating the F1-measure metrics in `metrics.py` and `__init__.py`.

- Finally, there are some minor changes like using `ruamel.yaml.safe_load()` instead of `ruamel.yaml.load()` since using the latter is no longer considered safe and raises warnings.

### DIFF
--- a/tests/test_preprocessing.py
+++ b/tests/test_preprocessing.py
@@ -255,13 +255,13 @@ def check_scaling_features(use_feature_hashing=False, use_scaling=False):
 
     # these are the expected values of the f-measures, sorted
     if not use_feature_hashing:
-        expected_fmeasures = ([0.77319587628865982, 0.78640776699029125] if
+        expected_fmeasures = ([0.55276381909547745, 0.55721393034825872] if
                               not use_scaling else
-                              [0.94930875576036866, 0.93989071038251359])
+                              [0.65217391304347827, 0.70370370370370372])
     else:
-        expected_fmeasures = ([0.42774566473988435, 0.5638766519823788] if
+        expected_fmeasures = ([0.54255319148936176, 0.59433962264150941] if
                               not use_scaling else
-                              [0.87323943661971837, 0.85561497326203206])
+                              [0.69950738916256161, 0.69035532994923865])
 
     assert_almost_equal(expected_fmeasures, fmeasures)
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - conda config --add channels desilinguist
   - conda update --yes conda
 install:
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION nomkl numpy scipy beautiful-soup six scikit-learn==0.17.1 joblib prettytable python-coveralls pyyaml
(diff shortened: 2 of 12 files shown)


tests/test_preprocessing.py
```diff
@@ -255,13 +255,13 @@ def check_scaling_features(use_feature_hashing=False, use_scaling=False):
 
     # these are the expected values of the f-measures, sorted
     if not use_feature_hashing:
-        expected_fmeasures = ([0.77319587628865982, 0.78640776699029125] if
+        expected_fmeasures = ([0.55276381909547745, 0.55721393034825872] if
                               not use_scaling else
-                              [0.94930875576036866, 0.93989071038251359])
+                              [0.65217391304347827, 0.70370370370370372])
```
dan-blanchard: These are pretty huge drops in f-measure. Should we be concerned about this? I honestly don't know.

desilinguist: Yeah. I was wondering about this myself. The changes in the other tests aren't as drastic but I am not sure what we can do about this.  Note also that the numbers look more reasonable to me now both with and without scaling and the direction is what we'd expect (scaling leads to better performance). Perhaps the previous implementation had bugs? 

@mheilman any thoughts?

desilinguist: Over the next few days, I am going to try and nail down what caused these differences. From what I can tell right now, It could be due to differences in:

- `make_classification_data`: perhaps the data being generated is actually different now?
- `SGDClassifier` - may be they changed something in the implementation?
- `FeatureHasher` - may be they changed something in the implementation?

desilinguist: It's looking like `make_classification()` generates different results in 0.18.1 than it used to with 0.17.1 even if we pass in the same random state. I have filed an [issue](https://github.com/scikit-learn/scikit-learn/issues/8180) with the scikit-learn folks. Let's see what they say.

desilinguist: Yeah, looks like they [changed](http://scikit-learn.org/stable/whats_new.html#enhancements) the underlying function `sample_without_replacement()` which causes the samples to be different for the same random state. So, I think we can safely say that the difference in the results are because of that.

desilinguist: @dmnapolitano can you finish your review when you have time?

desilinguist: Okay, to double check that this is indeed the cause of the extreme changes, I generated the test data in a conda environment with scikit-learn 0.17.1 and serialized it to disk. Then in a second conda environment with scikit-learn v0.18.1, I loaded this serialized data and ran this specific test. The numbers match. So, the differences are all because of changes to `make_classification()`.

dan-blanchard: Awesome. Thanks for doing that thorough investigation!

desilinguist:
This should NOT affect people's own experiments significantly unless they
are creating fake data using `make_experiment()`.

Yes, some minor changes should be expected with a new version release. If
people really want to reproduce older experiments exactly, they should be
using pinned conda environments.

Issue #221, opened 2014-12-30
automatic learning curves


mheilman:
It would be nice to have functionality to automatically run experiments with different size training sets in order to plot performance as a function of training sample size.

Perhaps this could be a separate experiment type (e.g., like "evaluate").  This probably does not make sense for cross-validation.

Perhaps there should be an option not to save all the models, etc., since there could be hundreds.

Possible configuration options:
- a list of sample sizes to consider (the default could be powers of 2 starting at 32 or 64)
- the number of replications per sample size (this could default to 1, but higher values would produce smoother learning curves)

(This was also briefly mentioned in the discussion of #212.)

desilinguist:
This should use the learning curves feature from scikit-learn: http://scikit-learn.org/stable/modules/generated/sklearn.learning_curve.learning_curve.html

desilinguist:
@langep, has there been any progress on this issue?

desilinguist:
I'd like to include this in the upcoming v1.3.

Here's how I am thinking about this:

1. Have a new experiment type called `learning_curves` which will compute the learning curve using scikit-learn's built-in [method](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.learning_curve.html#sklearn.model_selection.learning_curve) and save the output as a CSV in the `results` directory. It will also save the actual learning curve plot as a PNG in the `results` directory if `matplotlib` is available.
2. The models for the various training sizes will _not_be saved.
3. Users will be able to specify the various training sizes and the number of cross-validation iterations to be used for averaging. 

Since this will be a new feature, I'd like to solicit input from all of you: @dan-blanchard @dmnapolitano @mheilman @bndgyawali @mulhod @benbuleong @aoifecahill @aloukina @cml54.

Thanks!

desilinguist:
addressed by #332.

Issue #150, opened 2014-10-08
results should include training/testing sample sizes


mheilman:
It would be very convenient to include sample sizes in the results files and summaries.

PR #332, opened 2017-01-23
Add learning curves

desilinguist:
- Addresses #221.
- The way that this works is by having a new task type called `learning_curve`.  This essentially ties in to a new `learning_curve()` method in the `Learner` class which is adapted from the scikit-learn method `sklearn.model_selection.learning_curve()`.  The reason that I didn't just basically call the scikit-learn method directly is because that method works with estimator objects and raw feature arrays . We want to apply the whole SKLL pipeline (feature selection, transformation, hashing, etc.) that the user has specified when computing the learning curve results and so we need to use the SKLL API.
- The process of computing the curve is as follows: Only a training set is required. For each point on the leaning curve, the training set is split into two partitions 80/20. The learner is trained on the subset of the 80% corresponding to the point of the learning curve and then evaluated on the 20% partition. This is repeated multiple times (using multiple different 80/20 partitions) and the results are averaged. This gives us the score for each point in the training curve. The whole process it then repeated for each point on the curve. 
- I consider the learning curve task to be orthogonal to ablation and finding the right hyper-parameters. Therefore, ablation and grid search are not allowed. Just like for the cross-validation task, no models are saved for this task.
- Users can specify the various training set sizes and the number of 80/20 partitions for each point tin the curve (if they don't, there are reasonable defaults for both).
- Users can also specify the number of cross-validation iterations to be used for averaging the results for a given training set size.
- The output of the `learning_curve` task is a TSV file containing the training set size and the averaged scores for all combinations of featuresets, learners, and objectives. If `pandas` and `seaborn` are available, actual learning curves are generated as PNG files - one for each feature set. Each PNG file contains a faceted plot with objective functions on rows and learners on columns. Here's an example plot.
foo_example_iris

(*Note*: since grid search is disallowed, we don't really need to train the learner for each objective separately; we could simply train the learner once and then compute the scores using multiple functions. However, this doesn't fit into the current parallelization scheme that SKLL follows and so I didn't feel like changing that. The training jobs are run in parallel so it's not that big a deal anyway.)


desilinguist:
Reviewers, please actually test this PR on your machines (on multiple datasets, if possible) to make sure things work as expected. Thanks!

bndgyawali:
I ran on a dataset and got this nice plot. 
screen shot 2017-01-23 at 5 00 09 pm

desilinguist:
Thanks @bndgyawali! Did you use the default values for the learning curve CV folds? Do you mind sharing your config file here?

bndgyawali:
Here it is. I did use all the default values. 
```[General]
experiment_name=exp_name
task=learning_curve

[Input]
train_directory=path_to_feature_directory
featuresets = [['abc_feature'], ['def_feature']]
learners=['LinearRegression', 'LinearSVR', 'RandomForestRegressor', 'Lasso', 'Ridge', 'ElasticNet']
suffix=.csv
feature_scaling=both

[Tuning]
grid_search=true
objectives=[quadratic_weighted_kappa, linear_weighted_kappa, qwk_off_by_one]

[Output]
log=/tmp/skll_output
results=/tmp/skll_output```

**Comment by desilinguist** [comment]
From @bndgyawali, it looks like this works pretty well on other datasets and for larger number of learners and objectives. I am currently trying to get the coverage issue resolved which is a little annoying because we don't want to require `matplotlib`.

**Comment by bndgyawali** [comment]
You had mentioned that grid_search is disallowed, is it better to show warning message if `grid_search=true`?

**Comment by desilinguist** [comment]
@bndgyawali it does do that already. See [here](https://github.com/EducationalTestingService/skll/blob/feature/add-learning-curves/skll/experiments.py#L875).

**Comment by desilinguist** [comment]
So, the reason for the weird all-zero curve for `Lasso` in the plot @bndgyawali showed is because the default `alpha` for `Lasso` is 1.0 which, apparently, for his data is a very high penalty and the scores end up being all zero. However, if you set a different alpha via `fixed_parameters`, the curve is not all zeros anymore. 

Ideally, you would first do cross-validation experiments with grid search and/or ablation and get the right set of features and hyper-parameters for your model. Then, you would fix those in the config file for the learning curve and see the impact of the size of the training data.

**Comment by desilinguist** [comment]
I just realized that I forgot to update the documentation. Will do that next :)

**Comment by desilinguist** [comment]
Okay @bndgyawali @aloukina @dan-blanchard @benbuleong @cml54 @mheilman @aoifecahill @bwriordan @mulhod, this PR can now be considered complete including changes to the documentation. You guys can now start reviewing and testing it on your own datasets to make sure things work as expected.

Don't worry about the coverage decrease. It's only 0.003% :)

**Comment by desilinguist** [comment]
@dan-blanchard will you have time to look at this PR this week?

**Comment by mulhod** [comment]
call_henry_wer_bias_bias length wer

Unfortunately, if the training set sizes are a bit larger, the numbers get squished together on the same line in graph.

Unrelated to this example, I saw that the graph key is only on the first graph if you provide multiple learners and/or multiple objectives. This might be the intention, obviously, but I was thinking that maybe it would be better if it were not part of the first graph and was just at the top/sides/bottom when there are multiple graphs, if possible.

This is the config I used in the example above:
```
[General]
experiment_name = call_henry_wer_bias
task = learning_curve

[Input]
train_directory = ~/skll_learning_curves/call_henry_wer_bias/features
id_col = id
label_col = y
featuresets = [["length", "bias", "wer"]]
learners = ['RescaledSVR']
ids_to_floats = False
suffix = .jsonlines
feature_scaling = both
fixed_parameters = [{"C": 100.0, "gamma": 0.001}]

[Tuning]
min_feature_count = 1
feature_scaling = none
objectives = ["unweighted_kappa"]

[Output]
log = ~/skll_learning_curves/call_henry_wer_bias/log
results = ~/skll_learning_curves/call_henry_wer_bias/results
```

**Comment by desilinguist** [comment]
@mulhod thanks for looking at the PR! 

1. Good call about the x-tick tables. I can rotate them so so that they are easily visible.

2. Yes, the legend is only on the first graph intentionally. I did try creating the legend outside of the plotting area but it never really worked that well for me. I'll look at it again but if it doesn't work, I think putting it in the first plot is a decent compromise.

**Comment by mulhod** [comment]
The system I used to conduct the experiment I reported on above was Linux Ubuntu:
```
$ uname -a
Linux frigga 3.13.0-103-generic #150-Ubuntu SMP Thu Nov 24 10:34:17 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

File: tests/test_output.py [hunk 1]
```diff
@@ -163,8 +201,8 @@ def check_summary_score(use_feature_hashing=False):
                                      '{})').format(learner_name, result_score,
                                                    summary_score))
 
-    # We itereate over each model with an expected
-    # accuracy score. T est proves that the report
+    # We iterate over each model with an expected
+    # accuracy score. Test proves that the report
     # written out at least as a correct format for
```
mulhod: "as" is meant to be "has", right?

aloukina:
Hmm… the alignment looks legit from the first look – I’ll do more checks but it’s very exciting if we could do whole chapters!!!

aloukina:
Some metrics like r2 could be negative when model performance is really bad. The plots seem to set the min of y axis to 0.

aloukina:
If one supplies learning_curve_train_sizes = [1.0, 100.0, 200.0] this raises an error since the numbers are interpreted as floats but are not within (0, 1] range. This is the way this is handled in scikit-learn so we could leave it as is or be user-friendly and add a checking function to convert those to integers before passing them to the learning_curve functions?

aloukina:
fig
Same as Matt's comment above: When the number of different train sizes is reasonably large, x axis gets squished. The plot also shows what happens with negative r2.

desilinguist:
Thanks @aloukina! Very useful comments. Will incorporate.

skll/config.py
```diff
@@ -529,27 +569,27 @@ def _parse_config_file(config_path):
...
     if not all([objective in SCORERS for objective in grid_objectives]):
-        raise ValueError('Invalid grid objective function/s: {}'
+        raise ValueError('Invalid grid objective function(s): {}'
                          .format(grid_objectives))
 
     # check whether the right things are set for the given task
     if (task == 'evaluate' or task == 'predict') and not test_path:
         raise ValueError('The test set must be set when task is evaluate or '
                          'predict.')
-    if (task == 'cross_validate' or task == 'train') and test_path:
```
desilinguist: Because if the user set the `test_path`, may be they wanted to do `evaluate` rather than `cross_validate`  or `train` and forgot to change the task. The SKLL philosophy has always been to make no assumptions and let the user fix the config file lest they run a really long experiment they didn't really want to.

desilinguist:
My latest commits (a) automatically rotate the x-tick labels if any of the sizes is >= 1000 and (b) automatically generate the correct y-limits for the learning curves based on the metrics.

Note that if you use the  `mean_squared_error` metric for learning curve, the learning curve has negative values because scikit-learn essentially turns `mean_squared_error` into -1 * `mean_squared_error` internally so that it can be optimized just like any other function where higher is better. I will be submitting a PR that renames `mean_squared_error` to `neg_mean_squared_error` soon.

desilinguist:
Okay, I fixed the coverage issue. @aloukina and @mulhod, I am now waiting on you to test the changes and if everything looks good, I will merge.

mulhod:
call_henry_wer_bias_bias length wer

This looks good!

One nitpick: I understand why it's -1 to 1 for kappa on the y-axis, but is there a way not make the minimum y-axis value -1 if none of the values are below 0?

desilinguist:
Hmm, yeah I think we can do that. Let me see.

desilinguist:
Okay, I tweaked the plot generation code to hide unnecessary areas of the plot. Here's the new plot of your data, @mulhod. What do you think?
call_henry_wer_bias_bias length wer

PR #335, opened 2017-02-08
Switch to MKL backend

desilinguist:
The anaconda folks have fixed any issues with the MKL package that were causing builds to fail earlier in 2016 and had caused us to switch to forcing the use of `openblas` (via the `nomkl` meta-package). So, we are switching back to using MKL which would also allow us to have a single conda-recipe again.


desilinguist:
Reviewers, please check out the branch and run some of your own experiments with the branch to make sure MKL works for everyone.

benbuleong:
I ran a few experiments using both this branch and the master branch. The datasets were those used for our metaphor experiments. There are no changes at all and P,R,F,Acc and Objective score are exactly the same.

desilinguist:
Thanks @benbuleong, can you approve the PR?

PR #357, opened 2017-08-17
Feature/dummy classifier learner

Lguyogiro:
* Added sklearn's `DummyClassifier` (used for generating a simple baseline) to the list of supported classifiers.
* Updated Documentation accordingly
* Added a test that covers each of the three possible 'strategy' values ("stratified", "most_frequent", and "constant").

### DIFF
--- a/tests/test_classification.py
+++ b/tests/test_classification.py
@@ -20,7 +20,7 @@ from io import open
 from os.path import abspath, dirname, exists, join
 
 import numpy as np
-from nose.tools import eq_, assert_almost_equal, raises, assert_equal
+from nose.tools import eq_, assert_almost_equal, raises
 
 from skll.data import FeatureSet
 from skll.data.writers import NDJWriter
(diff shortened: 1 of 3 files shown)


desilinguist:
LGTM 👍 but let's merge this after the encrypted slack credentials one since I want to get that into `master` ASAP.

desilinguist:
Hmm, I thought I fixed the Slack integration to show approvals but I guess not.

tests/test_classification.py
```diff
@@ -210,6 +210,38 @@ def check_sparse_predict_sampler(use_feature_hashing=False):
...
+    )
+
+    # Ensure predictions are correct for all strategies.
+    correct = []
+    for model_args, expected_output in toy_data:
+        learner = Learner('DummyClassifier', model_kwargs=model_args)
+        learner.train(train_fs)
+        predictions = learner.predict(test_fs)
+        correct.append(np.array_equal(expected_output, predictions))
+    assert_equal(correct, [True, True, True])
```
desilinguist: I think the already imported `eq_` would work fine for this no? Do we really need `assert_equal` here?

Lguyogiro: No we don't. I've updated the test to use `eq_` and removed the `assert_equal` import

desilinguist:
@aoifecahill if you can approve this, I'll merge it in.

PR #360, opened 2017-08-18
Update to Python 3.6

desilinguist:
- Remove and update dependencies in both `travis.yml` as well as the conda recipe (`logutils` does not seem to be used anywhere, `pyyaml` should be `ruamel.yaml`, and `beautiful-soup` has been renamed to `beautifulsoup4`)
- Switch to using `conda-forge` for some of the packages that are not in `defaults` instead of building them and saving them to the `desilinguist` channel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_install:
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then export PATH=/home/travis/miniconda2/bin:$PATH; else export PATH=/home/travis/miniconda3/bin:$PATH; fi
   - conda update --yes conda
 install:
-  - conda install --yes --channel defaults --channel conda-forge python=$TRAVIS_PYTHON_VERSION numpy scipy beautifulsoup4 six scikit-learn==0.18.1 joblib prettytable python-coveralls pyyaml
+  - conda install --yes --channel defaults --channel conda-forge python=$TRAVIS_PYTHON_VERSION numpy scipy beautifulsoup4 six scikit-learn==0.18.1 joblib prettytable python-coveralls ruamel.yaml
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes --channel defaults configparser mock; fi
   - if [ ${WITH_PANDAS_AND_SEABORN} == "true" ]; then conda install --yes --channel defaults pandas seaborn; fi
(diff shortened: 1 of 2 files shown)


desilinguist:
FYI, I tested the `conda-recipe` and it works fine.

.travis.yml
```diff
@@ -19,13 +19,11 @@ before_install:
...
   - ./miniconda.sh -b
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then export PATH=/home/travis/miniconda2/bin:$PATH; else export PATH=/home/travis/miniconda3/bin:$PATH; fi
-  - conda config --add channels dan_blanchard
-  - conda config --add channels desilinguist
   - conda update --yes conda
 install:
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION numpy scipy beautiful-soup six scikit-learn==0.18.1 joblib prettytable python-coveralls pyyaml
-  - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes configparser logutils mock; fi
-  - if [ ${WITH_PANDAS_AND_SEABORN} == "true" ]; then conda install --yes pandas seaborn; fi
+  - conda install --yes --channel defaults --channel conda-forge python=$TRAVIS_PYTHON_VERSION numpy scipy beautifulsoup4 six scikit-learn==0.18.1 joblib prettytable python-coveralls ruamel.yaml
```
dan-blanchard: I think all the `--channel defaults` in here are unnecessary, and I'm not sure why you don't want to install conda-forge versions of `configparser`, `pandas`, or `seaborn` if they exist. conda-forge is basically always on a newer version than the defaults channel, and what's in there will eventually become the defaults channel.

desilinguist: @dan-blanchard it's not unnecessary since if you don't specify that, the scipy and numpy versions don't use `mkl`, they use `openblas` instead which is less efficient and slower (at least according to what I have read). It's also not always true that `conda-forge` is always on a newer version. They didn't have `scikit-learn` 0.18.1 for at least a couple of months even after `defaults` did.

desilinguist: I do see your point that may be we don't need the `--channel defaults` bits for the `configparser`, `seaborn` etc. but I just have them there so that it's explicit for someone reading it.

dan-blanchard: Ah, so it's all to make `defaults` take precedence. I didn't get that at first, since `defaults` is always included in the channel list. Comment withdrawn.

Issue #349, opened 2017-05-23
Weird logging issue when specifying `cv_folds_file`


desilinguist:
If you specify `cv_folds_file`, the content of the folds file get printed out in the log in place of where the number of folds would have been if `num_cv_folds` was specified.

desilinguist:
Addressed by #365.

PR #365, opened 2017-09-13
Bugfix/cv folds file logging

Lguyogiro:
* Addresses issue #349 . Previously, `cv_folds` was assumed to be an int. But if "cv_folds_file" is specified in the config, `cv_folds` will be a dictionary mapping ids to fold numbers and requires new code to correctly log the number of folds.
* Fixed by checking the type of `cv_folds` and, if it is not an int, logging the number of unique fold ids from the dict values.
* Added a cv folds file to `tests/train/` and `test_cv_folds_file_logging` to test suite


Lguyogiro:
I've pushed one more minor change: previously the experiment name (in the config file) for testing the cv folds file logging was "test_save_cv_folds". I've updated to reflect what it is actually testing.

desilinguist:
Since this is a very simple change, I am just going to merge it now without waiting for another approval.

Issue #363, opened 2017-08-27
specifying cross validation folds slows down experiments


aoifecahill:
Attached is a version of the titanic `cross_validate.cfg` that specifies the cross validation folds (as output by the original titanic `cross_validate.cfg` config). 

Here are the timings with and without specifying the fold ids file: 
```
Titanic_CV_family.csv+misc.csv+socioeconomic.csv+vitals.csv_RandomForestClassifier.results:Total Time: 0:01:26.131353
Titanic_CV_specify_folds_family.csv+misc.csv+socioeconomic.csv+vitals.csv_RandomForestClassifier.results:Total Time: 0:09:54.971322

Titanic_CV_family.csv+misc.csv+socioeconomic.csv+vitals.csv_DecisionTreeClassifier.results:Total Time: 0:00:02.022798
Titanic_CV_specify_folds_family.csv+misc.csv+socioeconomic.csv+vitals.csv_DecisionTreeClassifier.results:Total Time: 0:00:01.908257

Titanic_CV_family.csv+misc.csv+socioeconomic.csv+vitals.csv_SVC.results:Total Time: 0:00:32.489498
Titanic_CV_specify_folds_family.csv+misc.csv+socioeconomic.csv+vitals.csv_SVC.results:Total Time: 0:05:17.998272

Titanic_CV_family.csv+misc.csv+socioeconomic.csv+vitals.csv_MultinomialNB.results:Total Time: 0:00:01.921427
Titanic_CV_specify_folds_family.csv+misc.csv+socioeconomic.csv+vitals.csv_MultinomialNB.results:Total Time: 0:00:02.363040
```

Some of the experiments take much longer (e.g. SVC and Random Forrest), while others are similar. 

All of these timings are from experiments run in local mode. 

[cross_validate_specify_folds.cfg.txt](https://github.com/EducationalTestingService/skll/files/1254730/cross_validate_specify_folds.cfg.txt)

desilinguist:
This is something we should definitely fix in the next release. Thanks for the detailed report!

desilinguist:
As a first check, I factored out the [part](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L1497-L1523) of `learner.cross_validate()` that figures out the details of the folds into a separate function and then ran some timing experiments using the titanic data. As the screenshot below shows, although using a folds file is a little slower than using an integer (as expected), it's nowhere near the levels of slowness indicated by the timings reported above by @aoifecahill. 

[...]

PR #367, opened 2017-10-12
Speed up the use of an external folds file

desilinguist:
1. Fix the bug that was causing the inner grid-search loop of a cross-validation experiment to use a single job instead of the number specified via `grid_search_jobs`.

2. Rename the `cv_folds_file` config file option to `folds_file` and update the documentation to reflect that it is applicable to both the training and the cross-validation experiments.. 

3. Add a new config file option called `use_folds_file_for_grid_search` which controls whether the inner-loop grid-search uses the folds specified via `folds_file`. It's set to `True` by default which means. Setting it to `False` means that the inner loop uses the value of `grid_search_folds` (3 by default). This option is _only_ used in case of cross-validation experiments and is ignored otherwise. Add documentation for this option.

4. Also add a new keyword argument called `use_custom_folds_for_grid_search` to the `learner.cross_validate()` method. This flag by default sets `grid_search_folds` to be the same as `cv_folds` if custom folds are specified. It can be set to `False` to prevent this. This is required to handle the class when `cross_validate()` is directly called via the API, by passing the configuration parser.

5. Add a new comprehensive test to test the setting the values of `cv_folds` and `grid_search_folds` in the configuration file based on the above options.

6. Refactor tests to deal with the renaming of `cv_folds_file` to `folds_file`. 

7. Add documentation for the `save_cv_folds` option which we had forgotten to do when the option was added back in the day.

8. Minor re-indentation of some part of the code for readability.

(*Note*: See #363 for detailed timing experiments.)


desilinguist:
@aoifecahill @dan-blanchard @Lguyogiro @aloukina this is now ready for review! :)

skll/config.py
```diff
@@ -455,7 +455,7 @@ def _parse_config_file(config_path):
     random_folds = config.getboolean("Input", "random_folds")
     if random_folds:
         if folds_file:
-            logger.warning('Specifying folds_file overrides random_folds')
+            logger.warning('Specifying \"folds_file\" overrides \"random_folds\".')
```
dan-blanchard: Nitpick: No need for the slashes on these quotes, since it's `"` inside `'`.

doc/run_experiment.rst
```diff
@@ -411,18 +411,40 @@ random_folds *(Optional)*
...
+
+For the ``cross_validate`` task, this file will be used to define the outer
+cross-validation loop and, if :ref:`grid_search` is ``True``, also for the
+inner grid-search cross-validation loop. If the goal of specifiying the folds
+file is to ensure that the model does not learn to differentiate based on a confound:
+e.g. the data from the same person is always in the same fold, it makes sense to
+keep the same folds for both the outer and the inner cross-validation loops.
+
+However, sometimes the goal of specifying the folds file is simply for the
+purposes of comparison to another existing experiment or another context
```
aloukina: purpose?

doc/run_experiment.rst
```diff
@@ -669,6 +691,15 @@ grid_search_jobs *(Optional)*
...
 number of grid search folds.
 
+.. _use_folds_file_for_grid_search:
+
+use_folds_file_for_grid_search *(Optional)*
+"""""""""""""""""""""""""""""""""""""""""""
+
+Whether to use the specified :ref:`folds_file` for the inner grid-search
+cross-validation loop when :ref:`task` is set to ``cross_validate``. Ignored
+for all other tasks. Defaults to ``True``.
```
aloukina: Perhaps add a note that this will always be set to ``True`` for ``train``?

doc/run_experiment.rst
```diff
@@ -882,6 +913,17 @@ predictions.
     :ref:`log <log>`, :ref:`models <models>`, and
     :ref:`predictions <predictions>`.
 
+
+.. _save_cv_folds:
+
+save_cv_folds *(Optional)*
+""""""""""""""""""""""""""
+
```
aloukina: Which loop folds would these be?

tests/test_input.py
```diff
@@ -1008,12 +1012,167 @@ def test_config_parsing_relative_input_paths():
...
+    # note: `fold_mapping` refers to the dictionary version of the folds file
+
+    # task, cv_folds/folds_file, grid_search_folds, use_folds_file_for_grid_search -> cv_folds, grid_search_folds
+    # ('train', None, None, None) ->  (None, 3)
+    # ('train', None, None, True) ->  (None, 3)
+    # ('train', None, None, False) ->  (None, 3)
+    # ('train', None, 7, None) ->  (None, 7)
+    # ('train', None, 7, True) ->  (None, 7)
+    # ('train', None, 7, False) ->  (None, 7)
+    # ('train', 5, None, None) ->  (None, 3)
```
aloukina: Is there a warning in this case?

tests/test_input.py
```diff
@@ -1008,12 +1012,167 @@ def test_config_parsing_relative_input_paths():
...
+    # ('train', 5, None, True) ->  (None. 3)
```
aloukina: '.' -> ','

tests/test_input.py
```diff
@@ -1008,12 +1012,167 @@ def test_config_parsing_relative_input_paths():
...
+    # ('train', 5, None, False) ->   (None, 3)
+    # ('train', 5, 7, None) ->  (None, 7)
```
aloukina: There should be a warning too, I suppose?

tests/test_input.py
```diff
@@ -1008,12 +1012,167 @@ def test_config_parsing_relative_input_paths():
...
+    # ('train', 5, 7, True) ->  (None, 7)
+    # ('train', 5, 7, False) ->  (None, 7)
+    # ('train', 'train/folds_file_test.csv', None, None) ->  (None, fold_mapping)
+    # ('train', 'train/folds_file_test.csv', None, True) ->  (None, fold_mapping)
+    # ('train', 'train/folds_file_test.csv', None, False) ->  (None, fold_mapping)
+    # ('train', 'train/folds_file_test.csv', 7, None) ->  (None, fold_mapping)
```
aloukina: Warning?

tests/test_input.py
```diff
@@ -1008,12 +1012,167 @@ def test_config_parsing_relative_input_paths():
...
+    # ('train', 'train/folds_file_test.csv', 7, True) ->  (None, fold_mapping)
+    # ('train', 'train/folds_file_test.csv', 7, False) ->  (None, fold_mapping)
+    # ('cross_validate', None, None, None) ->  (10, 3)
+    # ('cross_validate', None, None, True) ->  (10, 3)
+    # ('cross_validate', None, None, False) ->  (10, 3)
+    # ('cross_validate', None, 7, None) ->  (10, 7)
+    # ('cross_validate', None, 7, True) ->  (10, 7)
```
aloukina: Warning?

tests/test_input.py
```diff
@@ -1008,12 +1012,167 @@ def test_config_parsing_relative_input_paths():
...
+    # ('cross_validate', None, 7, False) ->  (10, 7)
+    # ('cross_validate', 5, None, None) ->  (5, 3)
+    # ('cross_validate', 5, None, True) ->  (5, 3)
```
aloukina: I would think in this case the user will expect the same folds used for both loops so we should warn them?

tests/test_input.py
```diff
@@ -1008,12 +1012,167 @@ def test_config_parsing_relative_input_paths():
...
+    # ('cross_validate', 5, None, False) ->  (5, 3)
+    # ('cross_validate', 5, 7, None) ->  (5, 7)
+    # ('cross_validate', 5, 7, True) ->  (5, 7)
+    # ('cross_validate', 5, 7, False) ->  (5, 7)
+    # ('cross_validate', 'train/folds_file_test.csv', None, None) ->  (fold_mapping, fold_mapping)
+    # ('cross_validate', 'train/folds_file_test.csv', None, True) ->  (fold_mapping, fold_mapping)
+    # ('cross_validate', 'train/folds_file_test.csv', None, False) ->  (fold_mapping, 3)
+    # ('cross_validate', 'train/folds_file_test.csv', 7, None) ->  (fold_mapping, fold_mapping)
```
aloukina: Warning?

doc/run_experiment.rst
desilinguist: Whatever folds end up being used.

tests/test_input.py
desilinguist: Why would there need to be a warning?

tests/test_input.py
desilinguist: Again, not sure why we need a warning.

tests/test_input.py
desilinguist: Same?

tests/test_input.py
desilinguist: @aloukina if you look in the code, I added some new warnings in `config.py`. This is a test, so not sure why we need warnings here?

desilinguist: @aoifecahill here are [the](https://github.com/EducationalTestingService/skll/blob/fix/cv-folds-file-is-slower/skll/config.py#L632) [new](https://github.com/EducationalTestingService/skll/blob/fix/cv-folds-file-is-slower/skll/config.py#L635) [warnings](https://github.com/EducationalTestingService/skll/blob/fix/cv-folds-file-is-slower/skll/config.py#L644) that are already in the PR (each word is a link).

doc/run_experiment.rst
aloukina: So if I have a `cross-validate` task, would these be outer or inner loop?

tests/test_input.py
aloukina: You ask for 5 folds but your request is ignored since you asked in the wrong place and you get 3.

tests/test_input.py
aloukina: Because your request for 5 is silently ignored.

tests/test_input.py
aloukina: Same, your request for 7 is ignored.

tests/test_input.py
aloukina: You don't need them in test of course - it's just easier to see all places with warnings when you have all combinations written out with expected behaviour.
 Since you know the code better, you know where the warnings are already in place. For others we could just add an issue to add them in the future.

tests/test_input.py
desilinguist: The 5 doesn't matter here because that's for `cv_folds` which is not used for `train` at all.

tests/test_input.py
desilinguist: Yes, there's a warning for this already.

tests/test_input.py
desilinguist: I checked and I think I already have the warnings where I think they are really important.

tests/test_input.py
desilinguist: Yeah, this is an issue of not reading the documentation so I personally think this is a user problem. PEBKAC :-p

doc/run_experiment.rst
desilinguist: This is always the outer loop - the ones referred to by `cv_folds`.

skll/config.py
```diff
@@ -599,6 +616,34 @@ def _parse_config_file(config_path):
         raise ValueError('The models path should not be set when task is '
                          '{}.'.format(task))
 
+    # set the folds appropriately based on the task:
+    #  (a) if the task is `train` and if an external fold mapping is specified
+    #      then use that mapping for grid search instead of the value
+    #      contained in `grid_search_folds`.
+    #  (b) if the task is `cross_validate` and an external fold mapping is specified and the
+    #      then use that mapping for the outer CV loop. Depending on the value of
```
aoifecahill: typo

PR #368, opened 2017-10-16
Fix the example data creation scripts

desilinguist:
- Fix bug in `FeatureSet` iteration methods that would have failed if we were storing the feature values in non-sparse matrices (e.g., by setting `sparse=False` in `Reader` methods).
- Rename all data creation scripts to be consistent in their naming (`make_<name>_example_data.py`).
- Fix imports in those scripts to be consistent with scikit-learn v0.19.
- Fix indentation and remove unneeded imports.

### DIFF
--- a/doc/run_experiment.rst
+++ b/doc/run_experiment.rst
@@ -16,7 +16,7 @@ things work, do the following from the command prompt:
 .. code-block:: bash
 
     $ cd examples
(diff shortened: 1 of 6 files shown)


desilinguist:
To review this, please try generating the data for all three examples and running one of the example config files for each of those three examples.

PR #372, opened 2017-10-18
Streamline the results file for cross-validation task

desilinguist:
The `.results` file is produced for the `evaluate` and the `cross_validate` tasks. Currently, for the `cross_validate` task, the `.results` file (a) prints out the entire folds dictionary if `folds_file` is specified (b) includes details that aren't relevant to the experiment. 

In this branch, I streamline the results file in the following way:
- Include the path to the folds file when specified. This required the modification of the return tuple from  `_parse_config_file` to include the folds file path. 
- Do not include stratified folds information when a folds file is specified since it's not used in that context.
- Modify number of CV folds to be `<n> via folds file` if a folds file is specified.
- Modify grid search folds to be `<n> via folds file` if the grid search ends up using the folds file.
- Show the value of `use_folds_file_for_grid_search` when appropriate.
- Show grid search related information (`grid_search_folds` and `grid_objective`)  _only_ when we are actually doing grid search.

To test this, I also included an entirely new test in `test_output.py`. Additional changes in the branch include:
- Modify some additional tests to deal with the modifications to the `.results` and, hence, `.results.json` files. 
- Minor changes to docstrings and documentation.


skll/experiments.py
```diff
@@ -282,20 +282,28 @@ def _print_fancy_output(learner_result_dicts, output_file=sys.stdout):
...
     print('Task: {}'.format(lrd['task']), file=output_file)
+    if lrd['folds_file']:
+        print('Specified Folds File: {}'.format(lrd['folds_file']),
+              file=output_file)
     if lrd['task'] == 'cross_validate':
         print('Number of Folds: {}'.format(lrd['cv_folds']),
               file=output_file)
-        print('Stratified Folds: {}'.format(lrd['stratified_folds']),
-              file=output_file)
+        if not lrd['cv_folds'].endswith('folds file'):
```
mulhod: Is "folds file" supposed to be "folds_file" here? Can we just make it check if the string is equal to "folds file"/"folds_file" rather than using endswith or is there some reason that might not work?

skll/experiments.py
```diff
@@ -282,20 +282,28 @@ def _print_fancy_output(learner_result_dicts, output_file=sys.stdout):
...
-    print('Using Folds File: {}'.format(isinstance(lrd['cv_folds'], dict)),
-          file=output_file)
+    if lrd['grid_search']:
+        print('Grid Search Folds: {}'.format(lrd['grid_search_folds']),
+              file=output_file)
+        print('Grid Objective Function: {}'.format(lrd['grid_objective']),
+              file=output_file)
+    if (lrd['task'] == 'cross_validate' and
+        lrd['grid_search'] and
+        lrd['cv_folds'].endswith('folds file')):
```
mulhod: See comment above.

skll/experiments.py
desilinguist: Nope, it's trying to match the string on the right hand side of the colon in the results file which should say `3 via folds file` so it will always ends with "folds file" not "folds_file".

skll/experiments.py
desilinguist: See reply above.

skll/experiments.py
mulhod: Ok, I see. Misread that.

desilinguist: No worries. If you want to see some example output for various conditions, you can play around with `cross_validate.cfg` in the titanic example and look at the results files.

tests/test_output.py
```diff
@@ -170,7 +171,7 @@ def check_summary_score(use_feature_hashing=False):
             # the learner results dictionaries should have 29 rows,
```
aoifecahill: I think the comment needs to be updated with the new total number of rows.

PR #377, opened 2017-10-25
Add new learners (mostly regressors)

desilinguist:
1. Add the following new regressors:
    - `BayesianRidge`
    - `DummyRegressor`
    - `HuberRegressors`
    - `Lars`
    - `MLPRegressor`
    - `RANSACRegressor`
    - `TheilSenRegressor`

2. Add the following new classifiers:
    - `MLPClassifier`
    - `RidgeClassifier`

3. Add default parameter grids where appropriate for these new learners. Blank default parameter grids for learners that do not have any hyper-parameters that make sense to tune by default and the decision is better left to the user.

4. Update documentation. 
    - Add documentation for new learners.
    - Fix incomplete documentation for previously existing learners.
    - Add a note about empty default parameter grids.

5. Update tests. 
    - Add tests for new learners.
     - Refactor some old tests for consistency and ease of debugging.

### DIFF
--- a/skll/learner.py
+++ b/skll/learner.py
@@ -66,6 +66,7 @@ from sklearn.metrics import (accuracy_score,
                              precision_recall_fscore_support)
 from sklearn.naive_bayes import MultinomialNB
 from sklearn.neighbors import KNeighborsClassifier, KNeighborsRegressor
+from sklearn.neural_network import MLPClassifier, MLPRegressor
 from sklearn.preprocessing import StandardScaler
 from sklearn.svm import LinearSVC, SVC, LinearSVR, SVR
 from sklearn.tree import DecisionTreeClassifier, DecisionTreeRegressor
@@ -108,6 +109,14 @@ _DEFAULT_PARAM_GRIDS = {AdaBoostClassifier:
                         KNeighborsRegressor:
                         [{'n_neighbors': [1, 5, 10, 100],
                           'weights': ['uniform', 'distance']}],
+                        MLPClassifier:
+                        [{'activation': ['logistic', 'tanh', 'relu'],
+                          'alpha': [1e-4, 1e-3, 1e-3, 1e-1, 1, 10, 100, 1000],
+                          'learning_rate_init': [0.001, 0.01, 0.1]}],
+                        MLPRegressor:
+                        [{'activation': ['logistic', 'tanh', 'relu'],
+                          'alpha': [1e-4, 1e-3, 1e-3, 1e-1, 1, 10, 100, 1000],
+                          'learning_rate_init': [0.001, 0.01, 0.1]}],
                         MultinomialNB:
                         [{'alpha': [0.1, 0.25, 0.5, 0.75, 1.0]}],
                         Lars:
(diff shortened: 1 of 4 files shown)


skll/learner.py
```diff
@@ -72,24 +81,42 @@
...
+                        [{'epsilon': [1.05, 1.35, 1.5, 2.0, 2.5, 5.0],
+                          'alpha': [1e-4, 1e-3, 1e-3, 1e-1, 1, 10, 100, 1000]}],
                         KNeighborsClassifier:
                         [{'n_neighbors': [1, 5, 10, 100],
                           'weights': ['uniform', 'distance']}],
                         KNeighborsRegressor:
                         [{'n_neighbors': [1, 5, 10, 100],
                           'weights': ['uniform', 'distance']}],
+                        MLPClassifier:
+                        [{}],
```
Lguyogiro: I know searching over `hidden_layer_sizes` takes too long, but what about the other parameters like `activation`, `solver`, `learning_rate`, etc?

Lguyogiro: I know searching over `hidden_layer_sizes` takes too long, but what about the other parameters like `activation`, `solver`, `learning_rate`, etc?

tests/test_classification.py
```diff
@@ -178,15 +182,40 @@ def test_sparse_predict():
...
 
+def test_mlp_classification():
+    train_fs, test_fs = make_classification_data(num_examples=600,
+                                                 train_test_ratio=0.8,
+                                                 num_labels=3,
+                                                 num_features=5)
+
+    # train an AdaBoostRegressor on the training data and evalute on the
+    # testing data
+    learner = Learner('MLPClassifier', model_kwargs={'solver': 'lbfgs'})
```
Lguyogiro: Should this comment say "train an MLPClassifier on the training data and evaluate on the testing data" ?

skll/learner.py
desilinguist: I didn't include the other parameters either because:

(a) Many of the hyper-parameters are conditionally dependent on each other. For example, `learning_rate` is _only_ used when `solver` = `sgd`,  `learning_rate_init` is _only_ used when `solver` = `sgd` or `adam`, and `early_stopping` is _only_ used when `solver` = `sgd`. So, it's very different from other learners in that sense and there isn't an easy way to put together a `product` over all the parameters in the grid that will all be valid together.

(b) Many of the hyper-parameters shouldn't just be blindly iterated over. For example, the `lbfgs` is not great when you have a lot of data but when you have a small amount of data, it's better but still quite slow. Whether or not to use early stopping interacts with whether you have sufficient data in the first place. 

IMO, these decisions about which sets of hyper-parameters should be searched together depends on the experiment and the data and are best left to the user. However, if you have a suggestion for a default parameter grid that makes sense (without a lot of extra logic to deal with conditional dependence) then I am happy to try it out.

desilinguist: Hmm, an option could be to have `sgd` solver and a single hidden layer be the default for any MLP and then have a grid iterate over `activation`, `learning_rate`, and `learning_rate_init`. In that case, if you want to do something different, you will have to change the solver via `fixed_parameters` and then also specify your own set of parameters to search via `param_grids`. Thoughts? @Lguyogiro @aoifecahill @mulhod ?

desilinguist:
@aoifecahill @mulhod this is now ready for review. @Lguyogiro you can probably take another look since I added the MLP grid this time around.

skll/learner.py
```diff
@@ -44,20 +44,29 @@
...
+                                  Ridge,
+                                  RidgeClassifier,
+                                  SGDClassifier,
+                                  SGDRegressor,
+                                  TheilSenRegressor)
 from sklearn.linear_model.base import LinearModel
 from sklearn.metrics import (accuracy_score,
                              confusion_matrix,
-                             precision_recall_fscore_support,
-                             SCORERS)
```
aoifecahill: do we not need this any more?

desilinguist: It was imported but never used in that file so I removed it.

aoifecahill: Ah, makes sense, thanks!

PR #378, opened 2017-10-30
Do not do grid search if no param grids are provided

desilinguist:
- We might have a situation where the default parameter grid for a learner is empty (this is the case for some new regressors that have been added as part of PR #377). In that case, if the user also doesn't provide a parameter grid to search but `grid_search` is `True`, there's no point actually doing the grid search and we should just use the default hyperparameters.
- This PR adds a warning to that effect and turns off grid search in this situation.
- It also changes some of the logging messages to actually use the readable learner name instead of the fully qualified class name.
- It fixes a bug with the custom learner used in the test which came up because there's no reason to do grid search for that majority class learner.

### DIFF
--- a/skll/learner.py
+++ b/skll/learner.py
@@ -870,7 +870,7 @@ class Learner(object):
             # not supported
             raise ValueError(("{} is not supported by" +
                               " model_params with its current settings."
-                              ).format(self._model_type))
+                              ).format(self._model_type.__name__))
 
         return res, intercept
 
@@ -916,7 +916,7 @@ class Learner(object):
         default_param_grid = _find_default_param_grid(self._model_type)
         if default_param_grid is None:
             raise ValueError("%s is not a valid learner type." %
-                             (self._model_type,))
+                             (self._model_type.__name__,))
 
         estimator = self._model_type(**self._model_kwargs)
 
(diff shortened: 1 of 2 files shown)


desilinguist:
This PR is now ready to review (@aoifecahill, @mulhod, @jbiggsets, @benbuleong).

doc/run_experiment.rst
```diff
@@ -949,9 +949,9 @@ SVR
 
 .. note::
     Note that learners not listed here do not have any default
-    parameter grids in SKLL either because there are no hyper-parameters
-    to be tuned or because there are too many and decisions about
-    which ones to tune and how depend on the data being used for the
+    parameter grids in SKLL either because either there are no
+    hyper-parameters to tune or decisions about which parameters
+    to tune (and how) depend on the data being used for the
```
benbuleong: Would it be clearer to actually provide a brief direction where the user can do this tuning? If this  "best left up to the user" step is already documented in the user manual, I think it's fine. If not, perhaps cross reference the class/function where this manual tuning is can be done?

desilinguist: By "best left up to the user", I just meant that the user should provide his/her own `param_grids` in the config file.

benbuleong: Thanks for the clarification.

PR #380, opened 2017-10-31
Improve SKLL logging

desilinguist:
The goal of this PR is to improve how logging works in SKLL. The current situation is that many, if not all, of the warning messages are logged to the console but never stored in any of the log files. This is not great since a lot of people may not pay attention to the console messages. 

In an ideal situation, the solution would be quite simple. The way that `logging` works is that `logging.getLogger(<name>)` _always_ returns the same logger instance no matter where that call is made _as long as_ the call is within the same Python interpreter process. However, the problem is that SKLL jobs can also run on other machines in completely different Python processes. 

The solution in this PR works as follows:

1. I create a function called `get_skll_logger()` that returns a `Logger` instance that can log messages both to the `sys.stderr` as well as to a specified file. Repeated invocations of `get_skll_logger()` with the same name (and the same file path) will return the same `Logger` instance. 

2. `get_skll_logger()` is first called from `config.py` to create an experiment-level logger where messages that are relevant to configuration parsing are logged. It is called again from `experiments.py` at the top level to log messages (to the _same_ `Logger` instance) about the _whole_ experiment (not the individual jobs). This logger writes to the console as well as to a file named `<experiment_name>.log` that is automatically created in the `log` directory.

3. `get_skll_logger()` is called a third time from inside `_classify_featureset()`. This is a second `Logger `instance that is at the individual job level (where job = featureset + learner + objective). All of the messages from `learner.py`, `readers.py`, and `writers.py` are logged by this logger. This logger writes to the console as well as to the files named `<job_name>.log` that SKLL already creates in the `log` directory.

4. So, at the end of the experiment, there is (a) `<experiment_name>.log` containing messages pertaining to configuration parsing as well as the top-level experiment, and (b) multiple `<job_name>.log` files containing messages specific to each job.

**Notes**:

- To get `Learner`, `Reader`, and `Writer` to log to the same job log file, I had to modify them to accept loggers as keyword arguments. 

- Although the console output shows the logger names (`"experiment"` for the main logger and `<job_name>` for the job logger(s)), the log files themselves just show the time stamps, the logging level, and the actual message. This is because the names of the files themselves tell you what is what.

- I also had to pass in logger arguments to some top-level functions for things to work perfectly and make a one of the static methods be a regular method instead. 

- I changed the default logging level for SKLL to be `INFO` rather than warnings since the INFO messages are actually quite useful. The `-v` flag for `run_experiment` changes the logging level to include DEBUG messages as well. 

- I ran several tests both on and off the grid and the log files are created properly and with the right messages. However, I encourage you to run some tests (e.g., the SKLL examples) yourself.

- I added a couple of tests that look at the new log files.

### DIFF
--- a/doc/run_experiment.rst
+++ b/doc/run_experiment.rst
@@ -1138,7 +1138,7 @@ In addition to the above log files that are specific to each "job"
 in the configuration file), SKLL also produces a single, top level "experiment"
 log file with only ``EXPERIMENT`` as the prefix. While the job-level log files
 contain messages that pertain to the specific characteristics of the job, the
-experiment-level log file will contains logging messages that pertain to the
+experiment-level log file will contain logging messages that pertain to the
 overall experiment and configuration file. The messages in the log files are
 in the following format:
 
@@ -1146,7 +1146,7 @@ in the following format:
 
     TIMESTAMP - LEVEL - MSG
 
-where ``TIMESTAMP`` refers to the exact time when the messages was logged,
+where ``TIMESTAMP`` refers to the exact time when the message was logged,
 ``LEVEL`` refers to the level of the logging message (e.g., ``INFO``, ``WARNING``,
 etc.), and ``MSG`` is the actual content of the message. All of the messages
 are also printed to the console in addition to being saved in the job-level log
--- a/doc/run_experiment.rst
+++ b/doc/run_experiment.rst
@@ -1133,6 +1133,25 @@ the result, log, model, and prediction files will share the prefix
 ``EXPERIMENT_FEATURESET_LEARNER``. For backward-compatibility, the same
 applies when a single objective is specified using ``objective=x``.
 
+In addition to the above log files that are specific to each "job"
+(a specific combination of featuresets, learners, and objectives specified
+in the configuration file), SKLL also produces a single, top level "experiment"
+log file with only ``EXPERIMENT`` as the prefix. While the job-level log files
+contain messages that pertain to the specific characteristics of the job, the
+experiment-level log file will contains logging messages that pertain to the
+overall experiment and configuration file. The messages in the log files are
+in the following format:
+
+.. code-block:: bash
+
+    TIMESTAMP - LEVEL - MSG
+
+where ``TIMESTAMP`` refers to the exact time when the messages was logged,
+``LEVEL`` refers to the level of the logging message (e.g., ``INFO``, ``WARNING``,
+etc.), and ``MSG`` is the actual content of the message. All of the messages
+are also printed to the console in addition to being saved in the job-level log
+files and the experiment-level log file.
+
 For every experiment you run, there will also be a result summary file
 generated that is a tab-delimited file summarizing the results for each
 learner-featureset combination you have in your configuration file. It is named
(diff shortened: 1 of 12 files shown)


desilinguist:
@dan-blanchard I tagged you on this review because you are probably more of an expert at logging than I am so I really would like your input.

doc/run_experiment.rst
```diff
@@ -1133,6 +1133,25 @@ the result, log, model, and prediction files will share the prefix
 ``EXPERIMENT_FEATURESET_LEARNER``. For backward-compatibility, the same
 applies when a single objective is specified using ``objective=x``.
 
+In addition to the above log files that are specific to each "job"
+(a specific combination of featuresets, learners, and objectives specified
+in the configuration file), SKLL also produces a single, top level "experiment"
+log file with only ``EXPERIMENT`` as the prefix. While the job-level log files
+contain messages that pertain to the specific characteristics of the job, the
+experiment-level log file will contains logging messages that pertain to the
```
mulhod: contains --> contain

doc/run_experiment.rst
```diff
@@ -1133,6 +1133,25 @@ the result, log, model, and prediction files will share the prefix
...
+overall experiment and configuration file. The messages in the log files are
+in the following format:
+
+.. code-block:: bash
+
+    TIMESTAMP - LEVEL - MSG
+
+where ``TIMESTAMP`` refers to the exact time when the messages was logged,
```
mulhod: messages -> message

skll/data/readers.py
```diff
@@ -695,19 +696,24 @@ def safe_float(text, replace_dict=None):
                          floats. Anything not in the mapping will be kept the
                          same.
     :type replace_dict: dict from str to str
+    :param text: The Logger instance to use to log messages. Used instead of
```
mulhod: The documentation here and below should have `logger` instead of `text` as the parameter name.

skll/utilities/run_experiment.py
```diff
@@ -28,66 +28,76 @@ def main(argv=None):
     :type argv: list of str
     """
     # Get command line arguments
-    parser = argparse.ArgumentParser(
-        description="Runs the scikit-learn experiments in a given config file.\
-                     If Grid Map is installed, jobs will automatically be \
-                     created and run on a DRMAA-compatible cluster.",
-        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
-        conflict_handler='resolve')
+    parser = argparse.ArgumentParser(description='Runs the scikit-learn '
```
mulhod: Nitpick: To save on space, it might be slightly better to just import `ArgumentParser`/`ArgumentDefaultsHelpFormatter` directly.

skll/experiments.py
```diff
@@ -143,7 +144,7 @@ def _write_summary_file(result_json_paths, output_file, ablation=0):
     learner_result_dicts = []
     # Map from feature set names to all features in them
     all_features = defaultdict(set)
-    logger = logging.getLogger(__name__)
+    logger = get_skll_logger('experiment')
```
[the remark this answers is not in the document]
desilinguist: From the Logging Cookbook: 

> Multiple calls to logging.getLogger('someLogger') return a reference to the same logger object. This is true not only within the same module, but also across modules as long as it is in the same Python interpreter process. 

So, I don't think there should be any conflicts as long as you are using multiple `run_experiment` calls.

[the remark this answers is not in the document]
aoifecahill: 👍

PR #384, opened 2017-11-07
Decouple evaluation metrics from tuning objectives

desilinguist:
Right now, the only metrics computed for any SKLL experiment are the tuning objectives. However, one may want to compute additional evaluation metrics without tuning on them. This PR makes that possible. 

- For `evaluate` and `cross_validate` tasks, one can now specify a `metrics` list in the Output section and those will be computed and saved at the end of the results file under an  "Additional Evaluation Metrics" section. For `train` and `predict` tasks, `metrics` is ignored since it's not relevant. 

- For `learning_curve` tasks, it doesn't actually make sense to have `objectives` anyway so you can specify `metrics` instead. Using `objectives` is still supported for backward compatibility but you will get a deprecation warning. (*Note*: as an implementation detail, we still use the "objectives" variable internally because then we can piggyback on that to parallelize the various jobs.)

- The parsing and validation of `objectives` and `metrics` is now factored out into a separate function to avoid code duplication.

- Update documentation. 

- Add several new tests and update existing tests to deal with the new additional metrics being produced in the outputs.

### DIFF
--- a/skll/config.py
+++ b/skll/config.py
@@ -70,6 +70,7 @@ class SKLLConfigParser(configparser.ConfigParser):
                     'min_feature_count': '1',
                     'models': '',
                     'num_cv_folds': '10',
+                    'metrics': "[]",
                     'objectives': "['f1_score_micro']",
                     'objective': "f1_score_micro",
                     'param_grids': '[]',
@@ -109,6 +110,7 @@ class SKLLConfigParser(configparser.ConfigParser):
                                    'learning_curve_cv_folds_list': 'Input',
                                    'learning_curve_train_sizes': 'Input',
                                    'min_feature_count': 'Tuning',
+                                   'metrics': 'Output',
                                    'models': 'Output',
                                    'num_cv_folds': 'Input',
                                    'objectives': 'Tuning',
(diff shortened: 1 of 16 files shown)


skll/config.py
```diff
@@ -627,6 +622,37 @@ def _parse_config_file(config_path, log_level=logging.INFO):
     if task in ['cross_validate', 'learning_curve'] and model_path:
         raise ValueError('The models path should not be set when task is '
                          '{}.'.format(task))
+    if task == 'learning_curve':
+        if len(grid_objectives) > 0 and len(output_metrics) == 0:
+            logger.warning("The \"objectives\" option "
+                           "is deprecated for the learning_curve "
+                           "task and will not be supported "
+                           "starting with the next release; please "
```
aoifecahill: maybe "after" instead of "starting with"?

skll/config.py
```diff
@@ -627,6 +622,37 @@ def _parse_config_file(config_path, log_level=logging.INFO):
...
+            raise ValueError('The "metrics" option must be set when '
+                             'the task is "learning_curve".')
+        elif len(grid_objectives) > 0 and len(output_metrics) > 0:
+            logger.warning("Ignoring \"objectives\" for the learning_curve "
+                           "task since \"metrics\" is already specified.")
+            grid_objectives = []
+    elif task in ['evaluate', 'cross_validate']:
+        # for other appropriate tasks, if metrics and objectives have
+        # some overlaps - we will assume that the user meant to include
+        # use the metric for tuning _and_ evaluation, not just evaluation
```
aoifecahill: use -->

desilinguist:
@aoifecahill @mulhod @cml54 @benbuleong this PR is now ready for review.

Issue #385, opened 2017-11-08
Refactor Travis test plan to get around 50 minute job limit


desilinguist:
Travis seems to have a 50 minute hard limit on any job in any build and that includes VM boot up time. Now that we have so many tests, hitting that 50 minute job limit seems to be happening with more recent PRs. Here's a breakdown of how long each `test_X.py` file takes on my MacBook Pro.

File  | Info
-----|----
test_ablation | 8 tests in 35.305s
test_classification | 100 tests in 56.227s
test_cv | 7 tests in 7.324s
test_custom_learner | 3 tests in 1.154s
test_featureset | 81 tests in 337.709s
test_input | 123 tests in 0.646s
test_metrics | 210 tests in 0.449s
test_output | 27 tests in 239.037s
test_preprocessing | 7 tests in 1.394s
test_regression | 102 tests in 261.897s
test_utilities | 478 tests in 21.969s

Obviously, these numbers doesn't add up to 50 minutes because (a) I am not setting up the environment each time and (b) the MacBook is _way_ faster than the Ubuntu 14.04 VMs that Travis uses). However, we can still go by the relative ratios - it looks like `test_featureset` takes the most time, followed by `test_regression` and then `test_output`. So, may be we can have 3 separate jobs: one for `test_featureset`, one for `test_regression` and one for `test_output` + all the other files.

Thoughts?

desilinguist:
Addressed by #387.

PR #387, opened 2017-11-09
Refactor travis build

desilinguist:
Travis has a 50 minute hard limit on any job in any build and that includes VM boot up time. Now that we have so many tests, hitting that 50 minute job limit seems to be happening with more recent PRs. This PR breaks up the running of the test files into multiple jobs rather than running them all together. The solution implemented is detailed in #385.  

With this refactoring, the Travis CI plan runs for a total 26 minutes, taking into account the parallelization of the build jobs. And the longest running job is only ~20 min. See more details on [Travis CI](https://travis-ci.org/EducationalTestingService/skll).

This PR also solves some problems with `test_ablation.py`:
- Currently unless you run `test_ablation` first, some of the other tests will fail because they rely on a jsonlines file that is created by the ablation tests. I refactored the code such that the files are now created on demand even if `test_ablation` hasn't been run first.
- Create separate templates for all tests so that there are no strange conflicts when cleaning up.
- Improve readability.

### DIFF
--- a/tests/test_output.py
+++ b/tests/test_output.py
@@ -31,12 +31,18 @@ from sklearn.model_selection import ShuffleSplit, learning_curve
 from sklearn.naive_bayes import MultinomialNB
 
 from skll.data import FeatureSet, NDJWriter, Reader
-from skll.experiments import _HAVE_PANDAS, _HAVE_SEABORN, run_configuration, _compute_ylimits_for_featureset
+from skll.experiments import (_HAVE_PANDAS,
+                              _HAVE_SEABORN,
+                              _compute_ylimits_for_featureset,
+                              run_configuration)
 from skll.learner import Learner, _DEFAULT_PARAM_GRIDS
 
 from six import PY2
 
-from utils import fill_in_config_options, fill_in_config_paths, make_classification_data
+from utils import (create_jsonlines_feature_files,
+                   fill_in_config_options,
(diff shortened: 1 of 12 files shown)


desilinguist:
@mulhod @aoifecahill @Lguyogiro @jbiggsets this PR is now ready for review.

tests/test_output.py
```diff
@@ -52,6 +58,11 @@ def setup():
         if not exists(new_dir):
             os.makedirs(new_dir)
 
+    # create jsonlines feature files if they don't already exist
+    feature_files = [join(_my_dir, 'train', 'f{}.jsonlines'.format(i)) for i in range(5)]
+    if not all([exists(ff) for ff in feature_files]):
+        create_jsonlines_feature_files()
+
```
Lguyogiro: This code block is copied a number of times in a few different files (granted it is only 3 lines, but still that suggests to me it could be more modular). Would it make sense to put more of this code into the 'create_jsonlines_feature_files()' function? Or maybe just into another util function that checks if the files exist before calling 'create_jsonlines_feature_files()'. Then we could replace these three lines with a single function call eg "create_feature_files_if_none_exist()" or something similar.

desilinguist: Hmm, yeah may be `create_jsonlines_feature_files()` could check whether the files already exist and only then create them? I'll change that.

Issue #389, opened 2017-11-10
Improve y-limits for learning curves


desilinguist:
If we have learning curves for which the scores are very close to the maximum for the given metric (see attached example below), the curves have limits that theoretically may not make sense. It's not a big issue but perhaps we can fix it although we already have [some](https://github.com/EducationalTestingService/skll/blob/master/skll/experiments.py#L1195) intelligent computation of y-limits in SKLL that should be handling this.

![titanic_learning_curve_all](https://user-images.githubusercontent.com/161680/32637068-b35ae078-c585-11e7-9b28-0dffcfa682fa.png)

desilinguist:
I think I see what's going on. The y-limit computation function uses the following statement:

```
upper_limit = 1.1 if max_score <= 1 else math.ceil(max_score)
```

where `max_score` is computed as the sum of the mean plus the standard deviation. 

In the above example, the mean is 0.996401028277635 and the std. dev. is 0.011380947877212683 and the sum is 1.0077819761548477 for which the ceiling is 2 and hence the weird limits. We should be able to fix this.

desilinguist:
Addressed by #390.

Issue #386, opened 2017-11-09
Bug in SKLL when generating learning curves via gridmap


desilinguist:
Tested with the `learning_curve.cfg` file for the Titanic example included with SKLL. Running it locally works fine but running it on the grid yields the following error:

```
Traceback (most recent call last):
  File "/scratch/nmadnani/skllenv/bin/run_experiment", line 11, in <module>
    load_entry_point('skll', 'console_scripts', 'run_experiment')()
  File "/private/scratch/nmadnani/skll/skll/utilities/run_experiment.py", line 124, in main
    log_level=log_level)
  File "/private/scratch/nmadnani/skll/skll/experiments.py", line 1117, in run_configuration
    _check_job_results(job_results)
  File "/private/scratch/nmadnani/skll/skll/experiments.py", line 1154, in _check_job_results
    if not result_dicts or 'task' not in result_dicts[0]:
KeyError: 0
```

desilinguist:
Addressed by #390.

PR #390, opened 2017-11-10
Fix learning curve gridmap bug and tweak y-limit computation.

desilinguist:
This is a very short PR.

- Fix return statement in `_classify_featureset()` such that running a learning curve task via `gridmap` actually works (see issue #386).
- Tweak how the y-limits for learning curves are computed so that we don't have a lot of wasted space (see issue #389). The curve below shows the new version of the curve that was attached in the issue.
- Update the y-limits test based on the tweak.

titanic_learning_curve_all


desilinguist:
This PR is ready for review. You can click on the last green check mark in the commit history to see that the tests all passed and that the coverage is the same.

PR #392, opened 2017-11-13
Multiple minor bug fixes and improvements

desilinguist:
This is another short PR that encompasses several minor bugfixes and improvements:

1. Update to scikit-learn v0.19.1. In this particular scikit-learn release, the stratified split functions fail if the labels are floating point so we need to catch that condition in SKLL to provide a more informative message. Added such a condition to `learner.cross_validate()` and a test.

2. Deal with empty feature files by raising an exception. Added a test and also fixed an unbound variable bug.

3. Add `neg_log_loss` as a possible metric and tuning objective for classifiers. This requires `probability` to be `True` so add a check for that and a test.

4. If we are just specifying a `train_file` and a `test_file`, there are no featuresets and hence it doesn't make any sense to do ablation in that case. Added a warning and a fall-through to handle this scenario. 

5. Updated documentation for all of the above as appropriate.

### DIFF
--- a/skll/config.py
+++ b/skll/config.py
@@ -654,6 +654,12 @@ def _parse_config_file(config_path, log_level=logging.INFO):
                 output_metrics = [metric for metric in output_metrics
                                   if metric not in common_metrics_and_objectives]
 
+    # if the grid objectives contains `neg_log_loss`, then probability
+    # must be specified as true since that' needed to compute the loss
+    if 'neg_log_loss' in grid_objectives and not probability:
+        raise ValueError("The 'probability' option must be true in order "
+                         "to use `neg_log_loss` as the objective.")
+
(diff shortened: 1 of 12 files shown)


skll/config.py
```diff
@@ -654,6 +654,12 @@ def _parse_config_file(config_path, log_level=logging.INFO):
                 output_metrics = [metric for metric in output_metrics
                                   if metric not in common_metrics_and_objectives]
 
+    # if the grid objectives contains `neg_log_loss`, then probability
+    # must be specified as true since that' needed to compute the loss
```
mulhod: that' --> that's

desilinguist:
I fixed the coverage issue. This PR is now ready to be reviewed :)

aoifecahill:
Thanks!

Issue #373, opened 2017-10-18
Standardized docstrings to NumPy-style


jbiggsets:
This may be a low priority, but it would be useful to standardize all of the docstrings in `SKLL` to `numpy` style, and to make sure that all inputs and outputs are consistently documented.

There are a number of methods in the `SKLLConfigParser` class, for example, that do not have docstrings or whose docstrings do not describe the method's arguments and return values. These are all private methods, but it would still be useful for the purpose of code maintenance and PR reviews to document them.

desilinguist:
I'd appreciate if someone else could help tackle this issue.

desilinguist:
Addressed by #397.

PR #395, opened 2017-12-05
Feature/issue 344

Lguyogiro:
This PR updates `compute_eval_from_prediction` to support probabilities in the predictions file. Users can indicate if their predictions file contains probabilities with the `-p` flag. By default, the predicted class is that which has the highest probability. Additionally, calculating the prediction via the expected value when the classes are integers is supported. This is indicated with the argument `--prediction_method "expected_value"`.

I added two unit tests:
1. `test_compute_eval_from_predictions_with_probs` tests both the default behavior and the "expected_value" prediction method.
2. `test_compute_eval_from_predictions_breaks_with_expval_and_nonnumeric_classes` verifies that a `ValueError` is raised if the requested prediction method is "expected_value" but the classes are not integers.

Finally, in keeping with #373 ,  I updated the existing docstrings in `compute_eval_from_prediction`, as well as the docstring for the function I added, to NumPy style.


skll/utilities/compute_eval_from_predictions.py
```diff
@@ -10,25 +10,85 @@
 import argparse
 import csv
 import logging
+import random
```
aoifecahill: Should we set a random seed here too for reproducibility?

skll/utilities/compute_eval_from_predictions.py
```diff
@@ -78,6 +159,14 @@ def main(argv=None):
     parser.add_argument('metric_names',
                         help='metrics to compute',
                         nargs='+')
+    parser.add_argument('--probability', '-p',
+                        help="Whether the predictions file contains probabilities.",
+                        action='store_true')
+    parser.add_argument('--prediction_method', '-P',
+                        help="How to generate a prediction from the class probabilities "
+                             "(only use with -p flag). Supported methods: 'choose_best' "
```
aoifecahill: should we have an explicit check here that `--prediction_method` is not passed in without the `--probability` flag?

skll/utilities/compute_eval_from_predictions.py
```diff
@@ -39,7 +99,26 @@ def compute_eval_from_predictions(examples_file, predictions_file,
     pred = {}
     with open(predictions_file) as pred_file:
         reader = csv.reader(pred_file, dialect=csv.excel_tab)
-        next(reader)  # skip header
+        header = next(reader)
+
+        # If prediction file contains probabilities instead of predictions, convert
+        # class probabilities to a class prediction using the specified
+        # `prediction_mathod`.
```
aoifecahill: typo: `mathod` --> `method`

skll/utilities/compute_eval_from_predictions.py
```diff
@@ -10,25 +10,85 @@
...
+    classes: list
+        List of str or int class names.
+
+    probs: list of float
+        Probabilities for respective classes.
+
+    prediction_method: str
+        Indicates how to get a single class prediction from the probabilities. Currently
+        supported options are  "highest", which selects the class with the highest
+        probability, and "expected_value", which calculates an expected value over
```
desilinguist: What if two classes are tied for the highest probability? I believe in that case `max` returns the first one it encounters. This should be mentioned in the docstring.

skll/utilities/compute_eval_from_predictions.py
```diff
@@ -10,25 +10,85 @@
...
+        if len(set(probs)) == 1:  # probabilities are all equal
+            return random.choice(classes)
+        else:
+            return classes[probs.index(max(probs))]
+    elif prediction_method == 'expected_value':
+        exp_val = sum([classes[i] * prob for i, prob in enumerate(probs)])
+        return int(round(exp_val))
+
+
 def compute_eval_from_predictions(examples_file, predictions_file,
```
desilinguist: While we are adding a docstrings, can we fix the arguments to be one per line properly indented?

skll/utilities/compute_eval_from_predictions.py
```diff
@@ -10,25 +10,85 @@
...
+    examples_file: str
+        Path to a SKLL examples file (in .jsonlines or other format).
+
+    predictions_file: str
+        Path to a SKLL predictions output TSV file with id and prediction column names.
+
+    metric_names: list of str
+        A list of SKLL metric names (e.g., [pearson, unweighted_kappa]).
+
+    probabilities: boolean
```
desilinguist: I'd rename this keyword argument to be `has_probabilities`

skll/utilities/compute_eval_from_predictions.py
```diff
@@ -39,7 +99,26 @@ def compute_eval_from_predictions(examples_file, predictions_file,
...
-        next(reader)  # skip header
+        header = next(reader)
+
+        # If prediction file contains probabilities instead of predictions, convert
+        # class probabilities to a class prediction using the specified
+        # `prediction_mathod`.
+        if probabilities:
+            classes = [c for c in header[1:] if c]
+            if prediction_method == 'expected_value':
+                if any(not c.isdigit() for c in classes):
```
desilinguist: Why not just try to convert everything to `int` and catch a `TypeError` and then raise the exception?

skll/utilities/compute_eval_from_predictions.py
desilinguist: Yes, we should add that.

skll/utilities/compute_eval_from_predictions.py
```diff
@@ -78,6 +159,14 @@ def main(argv=None):
     parser.add_argument('metric_names',
                         help='metrics to compute',
                         nargs='+')
+    parser.add_argument('--probability', '-p',
+                        help="Whether the predictions file contains probabilities.",
+                        action='store_true')
+    parser.add_argument('--prediction_method', '-P',
```
desilinguist: I'd rename this argument to just be `method` and use `-m` to avoid confusion.

skll/utilities/compute_eval_from_predictions.py
desilinguist: Also, can't we tell that we have probabilities instead of the user informing us - based on the number of columns in the prediction file?

Lguyogiro: Yes we can do that. I wasn't sure which approach was preferable, and thought making the user specify would be simpler. I'll make this change.

skll/utilities/compute_eval_from_predictions.py
Lguyogiro: I will remove the `probability` flag, and generate a warning if there is a prediction method provided but the file does not contain probabilities.

desilinguist: And, in the other case - where we have probabilities but the user hasn't explicitly specified a method, we should print out a logging message saying we used the default prediction method.

skll/utilities/compute_eval_from_predictions.py
Lguyogiro: I considered both options, and favored asking permission over forgiveness, as I think it is easier to read. But I suppose trying to convert the classes to integers and catching the error is more robust.

skll/utilities/compute_eval_from_predictions.py
Lguyogiro: Maybe it is better to always make a random choice when two or more classes tie.

skll/utilities/compute_eval_from_predictions.py
```diff
@@ -39,7 +115,33 @@ def compute_eval_from_predictions(examples_file, predictions_file,
...
+                                                               probabilities,
+                                                               prediction_method)
+                pred[row[0]] = safe_float(prediction)
+        else:
+            if prediction_method is not None:
+                logger.warning("A prediction method was provided, but the predictions "
+                               "file doesn't contain probabilities. Ignoring prediction "
+                               "method '{}'.".format(prediction_method))
+
         for row in reader:
```
aoifecahill: Is there an indentation error here? Shouldn't this be nested underneath the `else` as part of the block that gets the predictions when they're not probabilities?

desilinguist: Good catch.

PR #409, opened 2018-01-26
Fix output directory creation

desilinguist:
- We had a bug in the code where `_locate_file()` was being called and we were depending on its return value to decide whether to create the output directories or not. However `_locate_file()` raises an exception when the file doesn't already exist, so this was a total waste of time.
- I fixed the code so that we actually catch the exception for non-existent output directories and then use that to create the directories if needed. 
- For the tests, I simply removed the pre-test manual creation of the `output` directory from `test_output.py` such that the automatic directory creation will be tested.

### DIFF
--- a/skll/config.py
+++ b/skll/config.py
@@ -481,10 +481,11 @@ def _parse_config_file(config_path, log_level=logging.INFO):
     # next, get the log path before anything else since we need to
     # save all logging messages to a log file in addition to displaying
     # them on the console
-    log_path = _locate_file(config.get("Output", "log"), config_dir)
-    if log_path:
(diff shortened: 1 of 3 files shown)


desilinguist:
Ugh, stupid coveralls! :(

desilinguist:
This is now ready for review. Please actually try an experiment on your local machines with directories to make sure.

desilinguist:
@benbuleong since you filled the original issue, can you please review this? @aoifecahill if you could test it too, it'd be great!

benbuleong:
Sure, I will review it by EOD

PR #427, opened 2018-05-18
Properly handle unlabeled data

Lguyogiro:
This PR makes Reader/Writers compatible with one another when the input file doesn't contain labels. 

- Updated `featureset.has_labels` to also return `False` when all of the labels are None (this is the case when reading a featureset with `label_col` set to `None`). 
- Added tests to make sure unlabelled data can be read by `CSVReader` and written with `NDJWriter`.


skll/data/featureset.py
```diff
@@ -414,8 +414,8 @@ def has_labels(self):
             Whether or not this FeatureSet has any finite labels.
         """
         if self.labels is not None:
-            return not (np.issubdtype(self.labels.dtype, float) and
-                        np.isnan(np.min(self.labels)))
+            return not ((np.issubdtype(self.labels.dtype, float) and np.isnan(np.min(self.labels))) or
+                        (all(label is None for label in self.labels)))
```
desilinguist: Might be better to write this new clause as `np.all(self.labels == True)`.

tests/test_featureset.py
```diff
@@ -972,3 +972,22 @@ def test_featureset_creation_from_dataframe_without_labels_with_vectorizer():
...
+    writer.write()
+    expected_path = os.path.join(_my_dir, "other",
+                                 "featureset_no_labels_expected.jsonlines")
+
+    with open(expected_path) as f_expected:
+        expected_output = f_expected.read()
+    with open(outpath) as f_actual:
+        actual_output = f_actual.read()
+
+    assert_equal(actual_output, expected_output)
```
desilinguist: Would `eq_` not work here?

Lguyogiro: It would, I'll update

desilinguist:
Keep it open.

aoifecahill:
keep it open

skll/data/featureset.py
desilinguist: Nope, I think the way you wrote it is actually recommended (and accurate).

PR #429, opened 2018-07-20
Allow gradient boosting models to take sparse matrices.

desilinguist:
- Remove `GradientBoostingClassifier` and `GradientBoostingRegressor` from list of learners that require dense matrices as input.


desilinguist:
The coverage decrease is expected.

desilinguist:
keep it open dude

PR #433, opened 2018-10-18
Issue 430/all probabilities get predictions

Lguyogiro:
This PR implements the feature suggested in PR-430. Namely, it adds an option in `generate_predictions` to output the probability for all labels, instead of just the indicated positive label. It also updates `generate_predictions` to behave more like the `Learner.predict` method in that it prints out the predictions in tsv format, optionally to a file. 

NOTE: This is will break backwards compatibility since we are now adding a header even in the single label case

### DIFF
--- a/skll/utilities/generate_predictions.py
+++ b/skll/utilities/generate_predictions.py
@@ -176,11 +176,6 @@ def main(argv=None):
                                 '%(message)s'))
     logger = logging.getLogger(__name__)
 
-    if args.positive_label and args.all_probabilities:
-        logger.warning("Ignoring `--positive_label` since "
-                       "`--all_probabilities` is set to True. The probability "
-                       "of all labels will be displayed.")
-
     # Create the classifier and load the model
     predictor = Predictor(args.model_file,
                           positive_label=args.positive_label,
--- a/skll/utilities/generate_predictions.py
+++ b/skll/utilities/generate_predictions.py
@@ -28,7 +28,7 @@ class Predictor(object):
     """
 
     def __init__(self, model_path, threshold=None, positive_label=1,
-                 return_all_probabilities=False, logger=None):
+                 all_labels=False, logger=None):
         """
         Initialize the predictor.
 
@@ -48,7 +48,7 @@ class Predictor(object):
             predicting. 1 = second class, which is default
             for binary classification.
             Defaults to 1.
-        return_all_probabilities: bool
+        all_labels: bool
             A flag indicating whether to return the probabilities for all
             labels in each row instead of just returning the probability of
             `positive_label`.
@@ -60,7 +60,7 @@ class Predictor(object):
         self._learner = Learner.from_file(model_path)
         self._pos_index = positive_label
         self.threshold = threshold
-        self.all_probs = return_all_probabilities
+        self.all_labels = all_labels
         self.output_file_header = None
 
     def predict(self, data):
@@ -84,7 +84,7 @@ class Predictor(object):
         # Create file header list, and transform predictions as needed
         # depending on the specified prediction arguments.
         if self._learner.probability:
-            if self.all_probs:
+            if self.all_labels:
                 self.output_file_header = ["id"] + [str(x) for x in labels]
             elif self.threshold is None:
                 label = self._learner.label_dict[self._pos_index]
@@ -122,29 +122,30 @@ def main(argv=None):
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
         conflict_handler='resolve')
     parser.add_argument('model_file',
-                        help='Model file to load and use for generating \
-                              predictions.')
-    parser.add_argument('input_file',
-                        help='A csv file, json file, or megam file \
-                              (with or without the label column), \
-                              with the appropriate suffix.',
+                        help='Model file to load and use for generating '
+                             'predictions.')
+    parser.add_argument('input_files',
+                        help='A space-separated list of csv file, json file, '
+                             'or megam file (with or without the label '
+                             'column), with the appropriate suffix.',
                         nargs='+')
     parser.add_argument('-i', '--id_col',
-                        help='Name of the column which contains the instance \
-                              IDs in ARFF, CSV, or TSV files.',
+                        help='Name of the column which contains the instance '
+                             'IDs in ARFF, CSV, or TSV files.',
                         default='id')
     parser.add_argument('-l', '--label_col',
-                        help='Name of the column which contains the labels\
-                              in ARFF, CSV, or TSV files. For ARFF files, this\
-                              must be the final column to count as the label.',
+                        help='Name of the column which contains the labels '
+                             'in ARFF, CSV, or TSV files. For ARFF files, '
+                             'this must be the final column to count as the '
+                             'label.',
                         default='y')
     parser.add_argument('-p', '--positive_label',
-                        help="If the model is only being used to predict the \
-                              probability of a particular label, this \
-                              specifies the index of the label we're \
-                              predicting. 1 = second label, which is default \
-                              for binary classification. Keep in mind that \
-                              labels are sorted lexicographically.",
+                        help="If the model is only being used to predict the "
+                             "probability of a particular label, this "
+                             "specifies the index of the label we're "
+                             "predicting. 1 = second label, which is default "
+                             "for binary classification. Keep in mind that "
+                             "labels are sorted lexicographically.",
                         default=1, type=int)
     parser.add_argument('-q', '--quiet',
                         help='Suppress printing of "Loading..." messages.',
@@ -180,7 +181,7 @@ def main(argv=None):
     predictor = Predictor(args.model_file,
                           positive_label=args.positive_label,
                           threshold=args.threshold,
-                          return_all_probabilities=args.all_probabilities,
+                          all_labels=args.all_probabilities,
                           logger=logger)
 
     # Iterate over all the specified input files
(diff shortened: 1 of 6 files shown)


skll/utilities/generate_predictions.py
```diff
@@ -26,7 +27,8 @@ class Predictor(object):
     predictions for feature strings.
     """
 
-    def __init__(self, model_path, threshold=None, positive_label=1, logger=None):
+    def __init__(self, model_path, threshold=None, positive_label=1,
+                 return_all_probabilities=False, logger=None):
```
desilinguist: A better keyword argument name might be `all_labels`:
```suggestion
                 all_labels=False, logger=None):
```

skll/utilities/generate_predictions.py
```diff
@@ -54,6 +60,8 @@ def __init__(self, model_path, threshold=None, positive_label=1, logger=None):
         self._learner = Learner.from_file(model_path)
         self._pos_index = positive_label
         self.threshold = threshold
+        self.all_probs = return_all_probabilities
```
desilinguist:
```suggestion
        self.all_labels = all_labels
```

skll/utilities/generate_predictions.py
```diff
@@ -71,18 +79,29 @@ def predict(self, data):
...
         preds = self._learner.predict(data)
         preds = preds.tolist()
+        labels = self._learner.label_list
 
+        # Create file header list, and transform predictions as needed
+        # depending on the specified prediction arguments.
         if self._learner.probability:
-            if self.threshold is None:
-                return [pred[self._pos_index] for pred in preds]
+            if self.all_probs:
```
desilinguist:
```suggestion
            if self.all_labels:
```

skll/utilities/generate_predictions.py
```diff
@@ -150,10 +180,11 @@ def main(argv=None):
     predictor = Predictor(args.model_file,
                           positive_label=args.positive_label,
                           threshold=args.threshold,
+                          return_all_probabilities=args.all_probabilities,
                           logger=logger)
 
     # Iterate over all the specified input files
-    for input_file in args.input_file:
+    for i, input_file in enumerate(args.input_file):
```
desilinguist: If `input_file` can be multiple files, let's just rename this to `input_files`?

skll/utilities/generate_predictions.py
```diff
@@ -169,8 +200,34 @@ def main(argv=None):
                                                     label_col=args.label_col,
                                                     id_col=args.id_col)
             feature_set = reader.read()
-            for pred in predictor.predict(feature_set):
-                print(pred)
+            preds = predictor.predict(feature_set)
+            header = predictor.output_file_header
+
+            if args.output_file is not None:
+                with open(args.output_file, "a") as fout:
```
desilinguist: I am not a fan of C-style file handle names:
```suggestion
                with open(args.output_file, "a") as outputfh:
```

tests/test_utilities.py
```diff
@@ -286,16 +286,17 @@ def test_compute_eval_from_predictions_random_choice():
...
+                               test_on_subset=False,
+                               all_probs=False):
 
     # create some simple classification feature sets for training and testing
     train_fs, test_fs = make_classification_data(num_examples=1000,
                                                  num_features=5,
                                                  use_feature_hashing=use_feature_hashing,
                                                  feature_bins=4)
-
+    proba = use_threshold or all_probs
```
desilinguist:
```suggestion
    enable_probabilities = use_threshold or all_probs
```

tests/test_utilities.py
```diff
@@ -351,8 +362,9 @@ def check_generate_predictions_console(use_threshold=False):
     writer = NDJWriter(input_file, test_fs)
     writer.write()
 
+    proba = use_threshold or all_probs
```
desilinguist:
```suggestion
    enable_probabilities = use_threshold or all_probs
```

tests/test_utilities.py
```diff
@@ -325,21 +326,31 @@ def check_generate_predictions(use_feature_hashing=False,
...
 
 
 def test_generate_predictions():
+    possibilities = [
+        (True, True, True, False), (True, True, False, False),
+        (True, False, True, False), (True, False, True, True),
+        (True, False, False, False), (True, False, False, True),
+        (False, True, True, False), (False, True, False, False),
+        (False, False, True, False), (False, False, True, True),
+        (False, False, False, False), (False, False, False, True)]
```
desilinguist: Why not just `product([True, False], [True, False], [True, False], [True, False])`?

tests/test_utilities.py
```diff
@@ -325,21 +326,31 @@ def check_generate_predictions(use_feature_hashing=False,
...
-         test_on_subset) in product([True, False], [True, False], [True, False]):
-        yield check_generate_predictions, use_feature_hashing, use_threshold, test_on_subset
+         test_on_subset,
+         all_probabilities) in possibilities:
+        yield (check_generate_predictions, use_feature_hashing,
+               use_threshold, test_on_subset, all_probabilities)
 
 
-def check_generate_predictions_console(use_threshold=False):
+def check_generate_predictions_console(use_threshold=False, all_probs=False):
```
desilinguist:
```suggestion
def check_generate_predictions_console(use_threshold=False, all_labels=False):
```

tests/test_utilities.py
Lguyogiro: Because `threshold` and `all_probabilities` are mutually exclusive. I could do the product and then filter out the cases I don't want to test, or add a test for those cases to verify that an error is raised, I suppose.

desilinguist: Yeah, the latter would a better option IMO.

desilinguist:
@Lguyogiro what’s happening with this PR? Do you need help with the coverage issue?

Lguyogiro:
@desilinguist Sorry I got a little sidetracked with something else. I will add tests and hopefully have the coverage issues resolved by end of week.

skll/utilities/generate_predictions.py
```diff
@@ -103,41 +126,53 @@ def main(argv=None):
...
     parser.add_argument('model_file',
-                        help='Model file to load and use for generating \
-                              predictions.')
-    parser.add_argument('input_file',
-                        help='A csv file, json file, or megam file \
-                              (with or without the label column), \
-                              with the appropriate suffix.',
+                        help='Model file to load and use for generating '
+                             'predictions.')
+    parser.add_argument('input_files',
```
mulhod: Just wondering, what's this change all about? It seems to be unrelated to the issue.

Lguyogiro:
This is now passing and the coverage drop has been fixed. Please feel free to review 👍

skll/utilities/generate_predictions.py
```diff
@@ -14,6 +14,7 @@
 import argparse
 import logging
 import os
+import sys
```
mulhod: Nitpick: It seems like this import is never used?

desilinguist:
@Lguyogiro actually you still need to update the PR per my original comment to say that this will break backwards compatibility.

skll/utilities/generate_predictions.py
```diff
@@ -46,14 +48,24 @@ def __init__(self, model_path, threshold=None, positive_label=1, logger=None):
             predicting. 1 = second class, which is default
             for binary classification.
             Defaults to 1.
+        all_labels: bool
+            A flag indicating whether to return the probabilities for all
+            labels in each row instead of just returning the probability of
+            `positive_label`.
```
jbiggsets: This is a nitpick, but can we add "Defaults to `False`" here?

skll/utilities/generate_predictions.py
```diff
@@ -169,8 +205,34 @@ def main(argv=None):
...
+                with open(args.output_file, 'a') as outputfh:
+                    if i == 0:  # Only write header once per set of input files
+                        print("\t".join(header), file=outputfh)
+                    if args.all_probabilities:
+                        for i, probabilities in enumerate(preds):
+                            id_ = feature_set.ids[i]
+                            probs_str = "\t".join([str(p) for p in probabilities])
+                            print("{}\t{}".format(id_, probs_str), file=outputfh)
+                    else:
+                        for i, pred in enumerate(preds):
```
jbiggsets: Can we call this `j` or something else, since we're using `i` in the outer loop? Same comment below.

Lguyogiro:
I've addressed @jbiggsets comments, and also added a unit test for the case of having multiple input files.

desilinguist:
I‘ll take a look at the other test you added on Monday.

skll/utilities/generate_predictions.py
```diff
@@ -71,18 +83,29 @@ def predict(self, data):
...
 
+        # Create file header list, and transform predictions as needed
+        # depending on the specified prediction arguments.
         if self._learner.probability:
-            if self.threshold is None:
-                return [pred[self._pos_index] for pred in preds]
+            if self.all_labels:
+                self.output_file_header = ["id"] + [str(x) for x in labels]
+            elif self.threshold is None:
+                label = self._learner.label_dict[self._pos_index]
```
slava92: This line seems to be a typo error. `self._pos_index` appears to be an index in an array but it is used here as a key in a dictionary. Was this supposed to be `label = self._learner.label_list[self._pos_index]` ?

desilinguist: You are right. Since `label_dict` maps string labels to ints, and `self._pos_index` is an int, we should be using `label_list` here not `label_dict`. I think we haven't run into this bug because the labels have also been ints for most of our problems? @aoifecahill @Lguyogiro any thoughts on this?

desilinguist: Let me move this into a separate issue. since this PR is actually closed.

Issue #434, opened 2018-11-29
MLP tests seem to be hanging


desilinguist:
Not sure what changed recently but @Lguyogiro reports that they hang for him locally too.

desilinguist:
What's interesting is that if I run `test_mlp_classification` on my Mac, it finishes in 122 seconds but on a Linux machine (a pretty beefy one at that). it takes 1771 seconds (almost half an hour). It feels like some underlying (unpinned) library changed that's causing the slowdown.

desilinguist:
Addressed by #435.

PR #435, opened 2018-11-29
Make MLP regression and classification tests run faster

desilinguist:
- Turn off grid search in MLP tests We need to do this since otherwise the travis builds time out. 
- This seems to be a problem on Linux machines (see #434).
- Turning off grid search is not ideal but it's not a terrible option.


Lguyogiro:
LGTM provided the pending travis build passes!

Issue #423, opened 2018-04-19
Explicitly document release process


desilinguist:
Just like RSMTool, we should have an explicitly documented release process so that other folks can do a  release if need be. Also make sure to add #422 to that document.

desilinguist:
Leave it open.

desilinguist:
Keep it open please.

desilinguist:
Addressed by #437.

Issue #422, opened 2018-04-19
Run `make linkcheck` as part of every release.


desilinguist:
We want to make sure that all external links are unbroken and sphinx includes a nice built-in way to do that. We should make sure to use it for every release.

desilinguist:
Leave it open.

desilinguist:
Keep it open please.

desilinguist:
Addressed by #437.

PR #437, opened 2018-11-30
Update documentation

desilinguist:
- Explicitly document the release process in the documentation (#423)
- Include `make linkcheck` as part of the release process (#422)
- Update all broken/redirected links

To test this branch, check it out locally and then go to the `doc` folder and run `make html`. Then look at the `_build/html/index.html` file. To check that the links are all fine, run `make linkcheck`. There are a couple of links that are still in `redirect` mode but I left them as-is for simplicity.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,5 +114,10 @@ documentation without the example gallery. The resulting HTML files will
 be placed in _build/html/ and are viewable in a web browser. See the
 README file in the doc/ directory for more information.
 
-For building the documentation, you will need [sphinx](http://sphinx.pocoo.org/).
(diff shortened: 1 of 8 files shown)


mulhod:
Worked for me!

One tangential comment: Can we add installation of `sphinx` or at least `sphinx_rtd_theme` to the instructions in `Contributing.md`. I installed `sphinx` with `conda` and then ran into the latter not being installed. Not really a big problem, but maybe worth it to add.

desilinguist:
Good call, @mulhod. I added the command `conda install sphinx sphinx_rtd_theme` to `Contributing.md` in the "Documentation" section.

Issue #414, opened 2018-02-11
Properly handle possible combinations of vectorizers in `Learner.predict()`


desilinguist:
I am not sure if it even makes sense to have a different number of feature bins in the training and test features sets. However, we do seem to allow for this possibility and try to handle it [here](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L1706). However, this code won't work because `FeatureHasher` objects do not have an `inverse_transform()` method. So, we should either just raise an exception if the non-matching feature bins doesn't make sense or try and actually fix the code.

desilinguist:
Still should do this I think.

desilinguist:
Keep it open.

desilinguist:
Keep it open.

desilinguist:
As I work on this issue, I think we need to think a bit more about this.

Let's enumerate the possible cases of the two vectorizers inside the `Learner.predict()` method - the vectorizer that was used for the feature set used to train the model and feature set that's been used to generate the feature set on which we want to make predictions.

1. Both vectorizers are `DictVectorizer`s. If they use different sets of features, we raise a warning and transform the features of the prediction set from its space to the trained model space. This is what we already do, is this case is fine.

2. Both vectorizers are `FeatureHashers`. If they use different number of feature bins, we should just raise an error since there's no `inverse_transform()` available for a `FeatureHasher` - the hash function is not reversible. 

3. The model vectorizer is a `FeatureHasher` but the prediction feature set vectorizer is a `DictVectorizer`. We should be able to handle this case, since we can just call `inverse_transform()` on the `DictVectorizer` and then `transform()` on the `FeatureHasher()`?  

4. The model vectorizer is a `DictVectorizer` but the prediction feature set vectorizer is a `FeatureHasher`. Again, we should raise an error here since there's no inverse available for the hasher. 

Does this logic look okay to everyone? @mulhod @Lguyogiro @jbiggsets?

desilinguist:
@mulhod any thoughts?

desilinguist:
Addressed by #445.

PR #445, opened 2018-12-12
Handle combination of model and data vectorizers in `Learner.predict()`

desilinguist:
This PR modifies `Learner.predict()` to handle the 4 combinations of possible vectorizer types - one for the model being used to make the prediction and the other for the data on which the predictions are to be made. It also adds new tests for previously unhandled combinations.

For a detailed discussion of this, please refer to the original issue #414.


skll/learner.py
```diff
@@ -1732,12 +1734,55 @@ def predict(self, examples, prediction_prefix=None, append=False,
...
+        both_hashers = model_is_hasher and data_is_hasher
+        model_hasher_and_data_dict = model_is_hasher and data_is_dict
+        model_dict_and_data_hasher = model_is_dict and data_is_hasher
+
+        # 1. both are DictVectorizers
+        if both_dicts:
+            if (set(self.feat_vectorizer.feature_names_) !=
+                    set(examples.vectorizer.feature_names_)):
                 self.logger.warning("There is mismatch between the training model "
                                     "features and the data passed to predict.")
```
jbiggsets: Should this warning also indicate how the mismatch will be resolved?

"The prediction features will be traansformed to the trained model space."

desilinguist: Good idea. I will do that!

Issue #452, opened 2019-02-05
`skll_convert` does not handle conversion of unlabeled data correctly in all cases


desilinguist:
For example, say, we have the following CSV file which does not have any labels.

```
id,feature1,feature2
1,0.2,0.8
2,0.1,0.9
3,0.99,0.01
4,0.0,1.0
```

And if I want to convert this to another format like ARFF, I have no way to tell `skll_convert` that this is unlabeled data. If I just run it:

```bash
$> skll_convert test.csv test.arff
```

the results are not what we would expect; `test.arff` now looks like this:

```
@relation 'skll_relation'

@attribute 'feature1' numeric
@attribute 'feature2' numeric
@attribute 'id' numeric

@data
0.2,0.8,1,0
0.1,0.9,2,0
0.99,0.01,3,0
0,1.0,4,0
```

This still has the labels at the end of the each data row which is not correct.

**Note**: this issue is related to #426 and was discovered while trying to address the latter.

desilinguist:
Addressed by #453 .

Issue #426, opened 2018-05-08
Reader/Writers not totally compatible for unlabelled feature sets


Lguyogiro:
I noticed that if you use a SKLL `Reader` to read an unlabelled `FeatureSet`, and then try to write that featureset with a SKLL `Writer`, an Exception is raised. This is due to incompatible handling of unlabelled data, where the Reader defines the labels as an array of `NoneType`, and the writer's `has_labels` checks if either the labels is `None` or the min of the labels array is `np.nan`. Since neither of these are true, it tries to write the labels as if they were floats, which causes a problem. 

To remedy this, either
1. The Reader should set the labels of an unlabelled `FeatureSet` to be an array of `np.nan`, or 
2. The `has_labels` property of `FeatureSet` should also check for `NoneType`.

To reproduce:
_data_ (as fake_data.csv):
```
id,feature1,feature2
1,0.2,0.8
2,0.1,0.9
3,0.99,0.01
4,0.0,1.0
```

_code_:
```
>>> from skll.data import CSVReader, NDJWriter
>>> fname = "fake_data.csv"
>>> reader = CSVReader(fname, label_col=None)
>>> fs = reader.read()
>>> writer = NDJWriter("out.jsonlines", fs)
>>> writer.write()
AttributeError: 'NoneType' object has no attribute 'item'
```

desilinguist:
@Lguyogiro what's happening with this PR?

Lguyogiro:
I ended up running into additional compatibility issues when testing all reader/writer combinations. It has been a few weeks, so when I have some time I will come back to this. Still planning to address the issue though, so let's leave this open.

aoifecahill:
keep it open please

desilinguist:
@Lguyogiro will you have time to do this for v1.5.3 release (which I am hoping to release in late January) or shall we push to v2.0 (which I imagine would come out in spring/summer)?

Lguyogiro:
I'll finish this before v1.5.3

desilinguist:
Great, thanks!

desilinguist:
We are going to move this to v2.0 since we need to do the v1.5.3 release soon.

desilinguist:
Addressed by #453.

PR #453, opened 2019-02-05
Properly handle unlabeled data in multiple places in SKLL

desilinguist:
1. Allow `skll_convert` to handle unlabaled input data (#452)
    - Add `--no_labels` option to `skll_convert`
    - Make `--no_labels` and `--label_col` mutually exclusive and add test for this.

2. Remove the `.ndj` format from the various conversion tests since it's identical to the `.jsonlines` format and just adds unnecessarily to the test run time.

3. Fix `FeatureSet.has_labels` to recognize list of `None` objects which is what happens when you read in an unlabeled data set and pass `label_col=None` (#426).

4. Fix bug in `ARFFWriter` that adds/removes `label_col` from the field names even if it's None to begin with.

5.  Update `test_convert_featureset()` in `test_featuresets.py` to also test for unlabeled data.


desilinguist:
@Lguyogiro @mulhod @jbiggsets any chance of reviewing this soon? I have another branch ready :)

desilinguist:
This is because we allow the CSV reader to [ignore non-existent columns and set the label to `None` if the column does not exist](https://github.com/EducationalTestingService/skll/blob/master/skll/data/readers.py#L643). Has nothing to do with conversion.

PR #458, opened 2019-02-11
Remove default value of objective function and `objective` config field.

desilinguist:
- Remove `objective` field and remove default values for `objectives.
- Remove code for normalizing and error checking the two fields since there is now only one.
- Add an additional check that `objectives` is not empty when running an experiment.
- Tweak parallelization to deal with empty objectives, e.g., when not doing grid search.
- Make default value of `grid_objective` to be `None` for `train()` and `cross_validate()` and add a check in `train()` that raises an exception if we are doing grid search and the objective is not specified. We don't need a similar check for `cross_validate()` since that calls `train()` underlyingly anyway.
- Make `metric` a required parameter for `learning_curve()`.
- Add new test for empty `objectives` and remove unnecessary tests e.g., for the  removed `objective` field.
- Fix tests that were using `objective` to now use `objectives`, explicitly specify functions where needed, and explicitly specify that we are doing grid search.
- Fix tests that were relying on `score` to not use that since that is the objective function value on the test set and will not be computed when there is no such function or when we aren't doing grid search.
- Ignore `grid_objectives` if we aren't doing grid search since this would prevent unnecessary computation when grid search is `False` but the user still specified a list of objectives.
- Add new tests.
- Flake8 formatting fixes in several places. 

Please test this quite throughly as this is a major backwards incompatible change.


desilinguist:
Something going on with Travis/Coveralls integration again. Re-running builds.

desilinguist:
Yeah, I don't really know why coveralls is being reported [3 new lines](https://coveralls.io/builds/21594954/source?filename=skll%2Flearner.py#L378) as being uncovered. I don't see an issue since that function is called multiple times in learning curve tests.

mulhod:
Even going back to the [last build](https://coveralls.io/builds/21576680/source?filename=skll%2Flearner.py#L322) with respect to `_train_and_score`, it seems like the function is not covered, at least according to Coveralls. I would guess that it claims this because when it's used it is called via `parallel(joblib.delayed(_train_and_score)...`.

desilinguist:
yeah that's what I was thinking as well @mulhod 👍

skll/config.py
```diff
@@ -390,7 +359,7 @@ def _parse_config_file(config_path, log_level=logging.INFO):
     do_grid_search : bool
         Whether to perform grid search.
     grid_objectives : list of str
-        A list of objects functions to use for tuning.
```
jbiggsets: Extremely nitpicky, but this should be "objective functions."

skll/config.py
```diff
@@ -990,7 +975,8 @@ def _parse_and_validate_metrics(metrics, option_name, logger=None):
     if not logger:
         logger = logging.getLogger(__name__)
 
-    # what are the objective functions for the grid search?
+    # make sure the given metrics data type is a list
+    # and parse it correctly
     metrics = yaml.safe_load(_fix_json(metrics))
     if not isinstance(metrics, list):
         raise TypeError("{} should be a list".format(option_name))
```
jbiggsets: Also nitpicky, but should we tell the user what the actual type is?

```
TypeError("{} should be a list, not {}.".format(option_name, type(metrics))
```

skll/config.py
desilinguist: Are you commenting on a line that was actually removed? :)

jbiggsets: Oh, my bad. I'm not sure how I did that! Please disregard. :)

PR #462, opened 2019-02-14
Write new tests to increase test coverage

desilinguist:
- Test for the `_train_and_score()` global function for both classifiers and regressors.
- Test for the `Learner.load()` method.
- Test for rescaling not applying to classifiers.
- Tests for custom learner API.
- Test for missing grid objective via Learner API.
- Removal of unnecessary imports and flake8 fixes.
- Tweaking the error message for when `grid_search` is `True` but there are no objectives specified via the API. Since `grid_search` is True by default, we need to indicate this to the user somehow since the default `learner.train(fs)` formulation will now fail. The older error message was not as informative.


tests/test_classification.py
```diff
@@ -671,3 +671,81 @@ def test_bad_xval_float_classes():
...
+    eq_(test_score1, test_score2)
+
+
+def test_train_and_score_function():
+    yield check_train_and_score_function, 'classifier'
+    yield check_train_and_score_function, 'regressor'
+
+
+@raises(ValueError)
+def test_learner_api_grid_search_no_objective():
```
Lguyogiro: It seems a little counter-intuitive to me to have two default values ('grid_search=True' and 'grid_objective=None') conflict. Is it not best practice for a method call with all default values to run successfully? I'm not sure if this would mean add a default grid_objective or turn off grid_search by default (I think I prefer the latter). I know this isn't the point of this PR but just a thought...

skll/learner.py
```diff
@@ -1386,7 +1386,7 @@ def train(self, examples, param_grid=None, grid_search_folds=3,
         if grid_search:
             if not grid_objective:
                 raise ValueError("You must specify a grid objective "
-                                 "if doing grid search.")
+                                 "or turn off grid search.")
```
Lguyogiro: IMHO this type of error message should contain two pieces: 1) a statement of what went wrong, followed by 2) the suggested action for fixing it...something like "grid_search set to True but no grid objective provided. You must either specify a grid objective or turn off grid search." I know I'm picking nits here.

tests/test_classification.py
desilinguist: Yes, that's a good point and I agree with you. Adding a default grid objective is not an option since that's what led to this in the first place (see PR #458). However, making `grid_search` have the default value of `False` would really break a _lot_ of things for people's code using this particular API the old way (including perhaps RSMTool). And the change would be silent - as in the code would run but just _without_ grid search. I was trying to minimize that.

skll/learner.py
desilinguist: Sure, that's a point well taken. I was trying to be succinct but I think perhaps being verbose is the better option here.

tests/test_classification.py
desilinguist: Based on discussion with the other project members, I think in this case we will incur the penalty of having the default values for `grid_search` and `grid_objective` conflict. For folks who were just doing `learner.train()` without specifying anything else, the error will be very early on in the process and they will now have to explicitly think about the metric that they were using which might be a good thing. Plus the code would not just run without grid search which could seriously impact downstream performance and that too, silently. 

A suggestion was made that maybe if `grid_search` is not explicitly specified, a warning can be shown that says it is being set to `True`. That way, in this case, where a user didn't realize that a grid search would be occurring (and didn't want it to), the warnings will let them know. However, this would mean we would have to come up with a way to distinguish between when `grid_search=True` is explicitly specified vs. when it's not specified but the value is still `True` since that's the default. This is complicated to say the least and I am not sure that the warning is worth the added complexity. 

We will make sure to explicitly mention this change in the release notes and highlight it in the documentation as well.

PR #465, opened 2019-02-15
Set default value of `grid_search` in config files to be `True`. 

desilinguist:
In order to be consistent with the API, it makes sense to set `grid_search` to be `True` by default in the config files. 

- Set default value of `grid_search` to be `True` in the config parser.
-  Update error message when no objectives are provided  to be more informative in light of this similar to the API error message.
- Move the grid search + learning curve check to config parsing instead of in `experiments.py` since it belongs here. And update the warning message to be more accurate.
- Update all tests and templates
    - If we were not previously specifiying grid search, we wanted it to be false so we need to be explicit about that now.
    - Explicitly specify grid search in API everywhere
- Fix some comments and docstrings in tests.
- Minor flake8 fixes.

### DIFF
--- a/tests/test_classification.py
+++ b/tests/test_classification.py
@@ -712,7 +712,7 @@ def test_train_and_score_function():
 
 
 @raises(ValueError)
-def test_learner_api_grid_search_no_objective():
+def check_learner_api_grid_search_no_objective(task='train'):
 
     (train_fs,
      test_fs) = make_classification_data(num_examples=500,
@@ -721,7 +721,15 @@ def test_learner_api_grid_search_no_objective():
                                          use_feature_hashing=False,
                                          non_negative=True)
     learner = Learner('LogisticRegression')
-    _ = learner.train(train_fs)
+    if task == 'train':
+        _ = learner.train(train_fs)
+    else:
+        _ = learner.cross_validate(train_fs)
+
+
+def test_learner_api_grid_search_no_objective():
+    yield check_learner_api_grid_search_no_objective, 'train'
+    yield check_learner_api_grid_search_no_objective, 'cross_validate'
 
 
 def test_learner_api_load_into_existing_instance():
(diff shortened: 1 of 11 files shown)


tests/test_input.py
```diff
@@ -532,8 +524,7 @@ def test_config_parsing_bad_test():
     test_dir = join(_my_dir, 'test')
     output_dir = join(_my_dir, 'output')
 
-    # make a simple config file that has a bad task
-    # but everything else is correct
+    # make a simple config file that has bad test paths
```
mulhod: Nitpick: make "paths" singular

tests/test_classification.py
```diff
@@ -644,6 +644,7 @@ def test_xval_float_classes_as_strings():
     prediction_prefix = join(_my_dir, 'output', 'float_class')
     learner = Learner('LogisticRegression')
     learner.cross_validate(float_class_fs,
+                           grid_search=True,
```
bndgyawali: do we need `grid_search=True` if it is true by default?

tests/test_cv.py
```diff
@@ -278,7 +278,6 @@ def test_retrieve_cv_folds():
                                                  stratified=False,
                                                  cv_folds=num_folds,
                                                  grid_search=False,
-                                                 grid_objective='f1_score_micro',
```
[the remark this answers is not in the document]
desilinguist: 1. We allow it because a lot of the time people want to run the same experiment just toggling grid search to false (when it was true before) and taking out the objectives would be an extra step that they would need to remember to do. We do [check this](https://github.com/EducationalTestingService/skll/blob/master/skll/config.py#L794) and raise a n appropriate warning and basically ignore the objectives going forward. Does that make sense?
2. Good catch on `cross_validate`. I will fix it to be consistent with train.

tests/test_classification.py
desilinguist: No we don't but I want people to start being explicit so it's good if our tests are too.

tests/test_cv.py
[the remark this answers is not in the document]
desilinguist: @mulhod please see my answer for 1 and review the latest changes for 2. And re-review. Thanks!

PR #466, opened 2019-02-21
Make `model_params` and `print_model_weights` work with hashed features

desilinguist:
- Modify `model_params` to work with hashed features. Since we lose the original feature names after hashing, we make up names in the format `hashed_feature_xx` in the output. 
- Tweak`print_model_weights` utility to print intercepts more clearly. For example, if the intercept list contains a single item, do not print it as a list.
- Update the utility function `make_regression_data()` to properly produce hashed feature data. This particular SKLL function relies on the sklearn function `make_regression()` which doesn’t know anything about hashed features. `make_regression()` works by generating random feature values, adding noise to them, generating weights randomly, and then computing the labels by taking dot product of the feature values and the weights. `make_regression_data()` then uses this data and creates train and test feature sets that are then used by tests. Previously when feature hashing was used, the learners in the tests were using the same labels as generated by `make_regresssion()` which doesn’t know anything about hashing. This meant that learners were using hashed versions of the feature values but the same labels so the linear relationship was not as strong. This meant lower correlation values in the tests. I now changed `make_regression_data()` to modify the labels based on the hashed versions of the features. This means that that the correlation values in the tests should now be > 0.9 range just like for unhashed features. 
- Update tests to deal with changes to `make_regression_data()`.
- Update regression and utility tests to cover the `model_params` and `print_model_weights` changes.

### DIFF
--- a/skll/learner.py
+++ b/skll/learner.py
@@ -1113,9 +1113,14 @@ class Learner(object):
             coef = self.model.coef_
             intercept = {'_intercept_': self.model.intercept_}
 
-            # convert SVR coefficient format (1 x matrix) to array
+            # convert SVR coefficient from a matrix to a 1D array
+            # and convert from sparse to dense also if necessary.
+            # However, this last bit may not be necessary
+            # if we did feature scaling and coef is already dense.
             if isinstance(self._model, SVR):
-                coef = coef.toarray()[0]
+                if sp.issparse(coef):
+                    coef = coef.toarray()
+                coef = coef[0]
 
             # inverse transform to get indices for before feature selection
             coef = coef.reshape(1, -1)
--- a/skll/learner.py
+++ b/skll/learner.py
@@ -14,13 +14,13 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 import copy
 import inspect
-import math
 import logging
 import os
 import sys
 from collections import Counter, defaultdict
 from functools import wraps
 from importlib import import_module
+from math import floor, log10
 from itertools import combinations
 from multiprocessing import cpu_count
 
@@ -1075,7 +1075,7 @@ class Learner(object):
             if isinstance(self.feat_vectorizer, FeatureHasher):
                 self.logger.warning("No feature names are available since this model was trained on hashed features.")
                 num_features = len(coef)
-                index_width_in_feature_name = math.floor(math.log10(num_features)) + 1
+                index_width_in_feature_name = int(floor(log10(num_features))) + 1
                 for idx in range(num_features):
                     if coef[idx]:
                         index_str = str(idx + 1).zfill(index_width_in_feature_name)
@@ -1104,7 +1104,7 @@ class Learner(object):
                 coef = self.feat_selector.inverse_transform(coef)[0]
                 if isinstance(self.feat_vectorizer, FeatureHasher):
                     num_features = len(coef)
-                    index_width_in_feature_name = math.floor(math.log10(num_features)) + 1
+                    index_width_in_feature_name = int(floor(log10(num_features))) + 1
                     for idx in range(num_features):
                         if coef[idx]:
                             index_str = str(idx + 1).zfill(index_width_in_feature_name)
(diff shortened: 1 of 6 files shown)


desilinguist:
FYI, please click on the green check mark on the penultimate commit to see the build results. The latest commit just fixes some comments in the code so I didn't want to re-run the CI build.

desilinguist:
Is it weird that coveralls reports the 93.54% coverage as an increase but for `master`, the coverage was actually [93.63%](https://coveralls.io/builds/21674155). So, coverage actually went down and not up, right?

desilinguist:
If possible, please try training a linear model with some hashed features and see what the output looks like as part of the review. Thanks!

mulhod:
I did some experimenting with the Boston data-set. I converted labels to classification labels in a naive way (`y` values less than 25 became "yes", above "no") and then trained an SVC with a linear kernel (and also I tried with LogisticRegression) both without feature hashing and with it (`hashed_features` set to 3) and I printed model weights. They look the way I would expect them to look. Below are the model weights that are printed out in both cases. (Not shown are the model weights when feature hashing was not used, but I did look at them as well and everything seemed to work.) I also tried training on the original data with `LinearRegression` and feature hashing.

```
# With LogisticRegression
$ print_model_weights examples/boston/output_classes_feature_hashing/Example_example_boston_LogisticRegression.model
2019-02-22 16:39:43,856 - skll.learner - WARNING - No feature names are available since this model was trained on hashed features.
== intercept values ==
0.008243059841  yes

Number of nonzero features: 3
-0.014798722941 yes     hashed_feature_1
0.006418599583  yes     hashed_feature_2
0.002487196267  yes     hashed_feature_3

# With SVC with a linear kernel
$ print_model_weights examples/boston/output_classes_feature_hashing/Example_example_boston_SVC.model
2019-02-22 16:41:40,369 - skll.learner - WARNING - No feature names are available since this model was trained on hashed features.
== intercept values ==
1.100812164307  no-vs-yes

Number of nonzero features: 3
-0.008244714004 no-vs-yes       hashed_feature_1
0.000798621987  no-vs-yes       hashed_feature_2
0.000533724744  no-vs-yes       hashed_feature_3
```

desilinguist:
Thank you so much for the comprehensive testing!

desilinguist:
Okay, @bndgyawali @jbiggsets @Lguyogiro, this PR is next. Please review this one when you get a chance.

skll/learner.py
```diff
@@ -1071,9 +1072,19 @@ def model_params(self):
             # inverse transform to get indices for before feature selection
             coef = coef.reshape(1, -1)
             coef = self.feat_selector.inverse_transform(coef)[0]
-            for feat, idx in iteritems(self.feat_vectorizer.vocabulary_):
-                if coef[idx]:
-                    res[feat] = coef[idx]
+            if isinstance(self.feat_vectorizer, FeatureHasher):
+                self.logger.warning("No feature names are available since this model was trained on hashed features.")
+                num_features = len(coef)
+                index_width_in_feature_name = int(floor(log10(num_features))) + 1
```
jbiggsets: Won't `floor()` always return an integer anyway?

desilinguist: That's what I had thought too but this is not the case for [Python 2](https://docs.python.org/2/library/math.html?highlight=math.floor#math.floor) which caused the python 2.7 tests to fail. Hence, the `int()` bit,.

skll/learner.py
```diff
@@ -1109,18 +1132,27 @@ def model_params(self):
...
                 coef = self.model.coef_[i]
                 coef = coef.toarray()
                 coef = self.feat_selector.inverse_transform(coef)[0]
                 class1 = self.label_list[class_pair[0]]
                 class2 = self.label_list[class_pair[1]]
-                for feat, idx in iteritems(self.feat_vectorizer.vocabulary_):
-                    if coef[idx]:
-                        res['{}-vs-{}\t{}'.format(class1, class2, feat)] = coef[idx]
-
+                if isinstance(self.feat_vectorizer, FeatureHasher):
```
jbiggsets: Not a big deal, but we're repeating this code a lot, and the only thing that's really changing is the actual feature name strings. Is there any way to goup this into a function, or is that too annoying?

desilinguist: Good call! I'll try and see if some refactoring is possible here.

desilinguist:
Thanks for the approvals! However, I am investing the loss of coverage and as part of that, realizing that we aren't testing for a few things. I will re-request reviews from both of you later today. Sorry for the inconvenience.

desilinguist:
The coverage decrease makes no sense because the last two commits should make _zero_ difference to coverage and the coverage for the [commit just before those two](https://coveralls.io/builds/21974839) was 93.71%!

PR #467, opened 2019-02-22
Replace prettytable

maxwell-schwartz:
This PR replaces instances of `prettytable`, which is no longer maintained, with `tabulate`.

### DIFF
--- a/skll/experiments.py
+++ b/skll/experiments.py
@@ -897,8 +897,9 @@ def _create_learner_result_dicts(task_results,
                 rows.append(result_row)
 
             result_table = tabulate(rows, 
-                                    headers=headers, 
-                                    floatfmt=".3f", 
+                                    headers=headers,
+                                    stralign="right",
+                                    floatfmt=".3f",
                                     tablefmt="grid",)
             result_table_str = '{}'.format(result_table)
             result_table_str += '\n(row = reference; column = predicted)'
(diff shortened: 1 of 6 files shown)


skll/experiments.py
```diff
@@ -897,8 +894,12 @@ def _create_learner_result_dicts(task_results,
                     f_sum_dict[actual_label] += float(label_f)
                 result_row = ([actual_label] + conf_matrix[i] +
                               [label_prec, label_recall, label_f])
-                result_table.add_row(result_row)
+                rows.append(result_row)
 
+            result_table = tabulate(rows, 
+                                    headers=headers, 
+                                    floatfmt=".3f", 
+                                    tablefmt="grid",)
```
desilinguist: - Please remove the extra whitespace after the commas.
- Add the keyword argument `stralign="right"` after `headers` to match the right alignment specification from PrettyTable.

skll/experiments.py
```diff
@@ -932,20 +933,19 @@ def _create_learner_result_dicts(task_results,
...
                 recall_mean = recall_sum_dict[actual_label] / num_folds
                 f_mean = f_sum_dict[actual_label] / num_folds
-                result_table.add_row([actual_label] +
-                                     [prec_mean, recall_mean, f_mean])
+                rows.append([actual_label] + [prec_mean, recall_mean, f_mean])
 
+            result_table = tabulate(rows, 
+                                    headers=headers, 
+                                    floatfmt=".3f", 
+                                    tablefmt="grid",)
```
desilinguist: - Again. please remove the extra whitespace after the commas.
- Change `grid` to `psql`  since that format better matches the last precision/recall table in the results file.

desilinguist:
I am still quite confused as to why the coverage decreased. Anyone have any ideas?

mulhod:
It seems to me that basically the total number of lines goes down by 1, so that would contribute to the issue a little bit (but not enough for a 0.1% drop, more like a 0.002% drop), even if the same number of lines were covered. I have some suspicions about how lines are being counted given the total numbers of counted lines and relevant lines reported in each build. It might be easier to talk this through in person.

desilinguist:
Let's not spend too much time on this coverage issue. If one of you can review this PR today, I will merge it and let's move on to the next one.

Issue #442, opened 2018-12-06
Improve `print_model_weights` output


desilinguist:
The output of `print_model_weights` sorts the non-zero weights by absolute value when printing them out. However, for multi-class classification with certain types of learners, it might make more sense to group the weights differently, e.g., by class for `LogisticRegression` and by class pair for SVC with linear kernels (which uses one-vs-one classification to do multi-class classification).

desilinguist:
Addressed by #468.

PR #468, opened 2019-02-22
improve print_model_weights output by adding a new argument to sort feature weights by sort_by_labels #442

bndgyawali:
This PR improves `print_model_weights` by adding a new argument `--sort_by_labels` which when true sorts the features by classes. By default, it is false. 

This new argument is mutually exclusive to `--k` which mean it only works when `--k` is not specified.


bndgyawali:
I am going to add a new test case. So, please hold a while to review it.

skll/utilities/print_model_weights.py
```diff
@@ -58,6 +64,13 @@ def main(argv=None):
     learner = Learner.from_file(args.model_file)
     (weights, intercept) = learner.model_params
 
+    multiclass = False
+    model = learner._model
+    if (isinstance(model, LinearSVC) or
+        isinstance(model, LogisticRegression) or
+        (isinstance(model, SVC) and
+            model.kernel == 'linear')):
+        multiclass = True
```
desilinguist: You could be using `LogisticRegression` for bianary classification in which case you will only have weights for one class and then you don't need to do this. May be you need an condition that the number of classes (via `self.label_list`) is  > 2?

skll/utilities/print_model_weights.py
```diff
@@ -69,19 +82,28 @@ def main(argv=None):
...
     print("Number of nonzero features:", len(weights), file=sys.stderr)
-    for feat, val in sorted(weight_items, key=lambda x: -abs(x[1]))[:k]:
-        print("{:.12f}\t{}".format(val, feat))
+    weight_by_class = defaultdict(dict)
+    if multiclass and args.sort_by_labels:
+        for label_feature, weight in weight_items:
+            label, feature = label_feature.split()
+            weight_by_class[label][feature] = weight
+        for label in sorted(weight_by_class):
+            for feat, val in sorted(weight_by_class[label].items(), key=lambda x: -abs(x[1])):
```
desilinguist: You need to show a warning to the suer that the `k` part is ignored when you are sorting.

bndgyawali:
I have addressed the comments and added extra tests to test the output of `print_model_weights` when `sort_by_labels` is passed as an argument. 
And, it's ready for review.

desilinguist:
@jbiggsets @mulhod if you get a chance, please review this PR.

PR #470, opened 2019-02-22
removing deprecated mean_squared_error, cv_folds_file and objectives in learning curve

bndgyawali:
This PR removes the support for `mean_squared_error`, `cv_folds_file` and `objectives` in learning curve.

`mean_squared_error` is replaced by `neg_mean_squared_error`
`cv_folds_file` is replaced by `folds_file`
`objectives` in learning curve is replaced by `metrics` in `output` section


skll/config.py
```diff
@@ -819,22 +811,18 @@ def _parse_config_file(config_path, log_level=logging.INFO):
...
     if task == 'learning_curve':
-        if len(grid_objectives) > 0 and len(output_metrics) == 0:
-            logger.warning("The \"objectives\" option "
-                           "is deprecated for the learning_curve "
-                           "task and will not be supported "
-                           "after the next release; please "
-                           "use the \"metrics\" option in the [Output] "
-                           "section instead.")
+        if len(grid_objectives) > 0:
+            raise ValueError("The \"objectives\" option "
```
desilinguist: I would rephrase this to make it a bit shorter:

"The \"objectives\" option is no longer supported for the learning_curve task. Please use the ..."

skll/config.py
```diff
@@ -988,20 +976,22 @@ def _parse_and_validate_metrics(metrics, option_name, logger=None):
...
-    # `mean_squared_error` should be replaced with `neg_mean_squared_error`
+    # `mean_squared_error` is no more supported.
+    # It is replaced by `neg_mean_squared_error`
     if 'mean_squared_error' in metrics:
-        logger.warning("The metric \"mean_squared_error\" "
-                       "is deprecated and will be removed in the next "
-                       "release, please use the metric "
-                       "\"neg_mean_squared_error\" instead.")
-        metrics[metrics.index('mean_squared_error')] = 'neg_mean_squared_error'
+        raise ValueError("The metric \"mean_squared_error\" "
```
desilinguist: You are repeating "is". 

Again, go shorter: 

"The metric \"mean_squared_error\" is no longer supported. Please use ..."

tests/configs/test_learning_curve_with_objectives.template.cfg
```diff
@@ -9,6 +9,6 @@ suffix=.jsonlines
 
 [Tuning]
 grid_search=false
-objectives=['accuracy', 'f1_score_micro']
 
 [Output]
+metrics=['accuracy', 'f1_score_micro']
```
desilinguist: I think you need to rename the template file since it no longer contains objectives.

tests/test_input.py
```diff
@@ -1715,7 +1753,7 @@ def test_learning_curve_metrics_and_no_objectives():
     eq_(grid_objectives, [])
```
desilinguist: You don't need this `eq_()` statement then?

tests/test_input.py
```diff
@@ -1751,7 +1789,7 @@ def test_learning_curve_objectives_and_no_metrics():
...
```
desilinguist: What about this `eq_()`

tests/test_input.py
```diff
@@ -1751,7 +1789,7 @@ def test_learning_curve_objectives_and_no_metrics():
...
 
 
-def test_learning_curve_default_objectives_and_no_metrics():
+def test_learning_curve_no_metrics():
```
desilinguist: Shouldn't this test raise an error? Is there a default learning curve metric?

bndgyawali: This test is calling `check_config_parsing_value_error` which raises `ValueError`. There is no default metrics. Instead, I can change this test case to call `_parse_config_file` and raise exception if you think that would be better.

tests/test_input.py
bndgyawali: removed.

skll/config.py
bndgyawali: fixed.

skll/config.py
bndgyawali: fixed.

tests/configs/test_learning_curve_with_objectives.template.cfg
bndgyawali: this template was duplicate of another, so it is now deleted.

tests/test_input.py
bndgyawali: removed

tests/test_input.py
```diff
@@ -1808,7 +1843,7 @@ def test_learning_curve_no_metrics():
                                          values_to_fill_dict,
                                          'learning_curve_default_objectives_and_no_metrics')
 
-    yield check_config_parsing_value_error, config_path
```
desilinguist: Why did you remove the `yield` here?

tests/test_output.py
```diff
@@ -456,6 +456,7 @@ def check_grid_search_cv_results(task, do_grid_search):
     # off, so simply turn it off here as well since it should
     # result in the same situation
     elif task == 'learning_curve':
+        values_to_fill_dict['metrics'] = values_to_fill_dict.pop('objectives')
```
desilinguist: Why is this needed?

bndgyawali: because, `learning_curve` does not support `objectives` but instead will need `metrics`.

tests/test_input.py
bndgyawali: I think I deleted it :) lemme put yield back.

tests/test_output.py
desilinguist: Yeah but I thought any tasks other than train and predict raise a ValueError anyway (see line 476 below)

bndgyawali: that is only for `train` and `predict`.

bndgyawali: I mean, ValueError is raised only if it is `train` and `predict` and `do_grid_search` is False. For the rest tasks, it will run `run_configuration(config_path, quiet=True)` in line 481

desilinguist: I see 👍

desilinguist:
Do we really need [these two lines](https://coveralls.io/builds/22184244/source?filename=skll%2Fconfig.py#L826) if we are raising a ValueError anyway. I think that's also why the coverage is going down.

Issue #431, opened 2018-09-21
Return cv_results_ results in addition to best_score_ from grid_searcher


rgokul:
For cross_validate task where we tune hyper parameters, often it is interesting to see plots of the performance of different hyper parameters on the train data. However, this is not possible because train() returns only the best score and the corresponding params.

I have made minor edits to dump it to a file for my needs but I would like to see this feature officially supported.

desilinguist:
Interesting idea. Thanks!

desilinguist:
Addressed by #471.

PR #471, opened 2019-02-22
Add grid search cv_results to results for train/cross_validate tasks

mulhod:
This change adds the `cv_results_` attribute of `GridSearchCV` (after fitting) to the return values of `Learner.train` and `Learner.cross_validate`. For `Learner.train`, a single `cv_results` value is returned (along with the best grid search score) since `GridSearchCV` was run one time (if `grid_search` is on, that is -- if not, the value is `None`). For `Learner.cross_validate`, a list of `cv_results_` dictionaries is returned alongside a list of best grid search scores, etc.

The `cv_results_` are put into the experiment's results JSON, in addition. Since the `train` and `predict` tasks previously did not allow one to specify a results directory, that has now been changed. This directory will only be written to if `grid_search` is on; otherwise, it will be ignored. In the case where the task is `train`/`predict`, the results JSON files that are created will contain only a single dictionary (since it is only run once, i.e., using all of the training data provided) with a key `"grid_search_cv_results"` mapped to the value of `cv_results_` dictionary in addition to other keys/values, including the best grid score, the experiment name, etc., just as it is in the `cross_validate`/`evaluate` tasks' results. In the `cross_validate`/`evaluate` case, this value will be inserted alongside `"grid_score"` whenever appropriate.

There are a couple minor changes I made that are unrelated or only slightly related to the issue, including spacing fixes, removal of unused variables, and more specific return value unpacking.


mulhod:
This PR addressed #431

skll/config.py
```diff
@@ -684,8 +684,6 @@ def _parse_config_file(config_path, log_level=logging.INFO):
     # make sure all the specified paths/files exist
     train_path = _locate_file(train_path, config_dir)
     test_path = _locate_file(test_path, config_dir)
-    train_file = _locate_file(train_file, config_dir)
```
desilinguist: Are these not used?

skll/experiments.py
```diff
@@ -733,12 +736,14 @@ def _classify_featureset(args):
...
             grid_scores = [None]
+            grid_search_cv_results_dicts = [None]
 
-        # print out the tuned parameters and best CV score
-        param_out = ('{}: {}'.format(param_name, param_value)
-                     for param_name, param_value in
-                     iteritems(learner.model.get_params()))
-        logger.info("Hyperparameters: {}".format(', '.join(param_out)))
+        # print out the tuned parameters
+        if grid_search:
```
desilinguist: Is the info not useful when we are not doing grid search?

skll/experiments.py
```diff
@@ -799,6 +805,20 @@ def _classify_featureset(args):
         with open(results_json_path, file_mode) as json_file:
             json.dump(res, json_file, cls=NumpyTypeEncoder)
     else:
+        if results_path:
+            results_json_path = join(results_path,
+                                     '{}.grid_search_cv_results.json'.format(job_name))
```
desilinguist: I'd argue that the results file should be named like all the other results files - using the job name rather than this fixed filename. Is there a reason for not doing that?

skll/config.py
mulhod: The variables `train_file` and `test_file` are not used again. This is because they get set to `train_path`/`test_path` right above this point (and those are the values that get returned - it's the logic around standardizing the file vs. directory issue). However, I see now upon closer inspection that `_locate_file` is checking the existence of the file, thereby catching that issue here and not later. So, these lines actually should not be removed. I think they can just be replaced with the method calls by themselves, though, since the return value is not used.

mulhod: Actually, the lines right above (for `train_path`/`test_path`) already do the checking. `train_path`'s value gets set to `train_file`'s value right above this, so, it's just duplicating it.

skll/experiments.py
mulhod: My thought was that the comment didn't really reflect what was happening since the hyperparameters will get get printed out whether they're tuned or not. Let me get rid of the conditional and just change comment, though. I can see why you'd want to know what parameters were used even if they're not being tuned via grid search.

skll/experiments.py
```diff
@@ -799,6 +804,20 @@ def _classify_featureset(args):
         with open(results_json_path, file_mode) as json_file:
             json.dump(res, json_file, cls=NumpyTypeEncoder)
     else:
+        if results_path:
```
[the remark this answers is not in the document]
desilinguist: I just tried running this. If we aren't doing `grid_search` but `results` is still set in the `Output` section of the config file, we will still write out a results file. Is that useful? Shouldn't we _only_ write out the results file for the `train` task if we are doing grid search AND `results` is set?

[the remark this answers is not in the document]
mulhod: Yeah, good point. Actually, I forgot about `predict`. So, the `else` will lead to the task being either `train` or `predict`. Both of these cases are similar with respect to what's happening here, though. If you specify a results path, then the grid search results will be written here just as they would be for `train`. I'll add a comment.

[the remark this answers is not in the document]
mulhod: No, it probably is not useful if grid search is not on. Let me take a look at what can be done here.

mulhod:
I've made some changes.

1. `train`/`predict` will fail if a results path is provided and grid search is turned off, just to avoid any confusion.
2. I've fleshed out the tests to cover all tasks and grid search on/off. The test files that are used in the tests give a pretty good indication of what would happen in each scenario. For example, the files `test_grid_search_cv_results_cross_validate_{gs,nogs}_f0_{SVC,LogisticRegression}.results.json` contain results for all combinations of grid search/no grid search and two learner types for the `cross_validate` task. (For `predict`/`train`, there are no corresponding "no grid search" files since it's not a valid experiment.)

skll/config.py
```diff
@@ -807,9 +805,9 @@ def _parse_config_file(config_path, log_level=logging.INFO):
     if task in ['cross_validate', 'train', 'learning_curve'] and test_path:
         raise ValueError('The test set should not be set when task is '
                          '{}.'.format(task))
-    if task in ['train', 'predict'] and results_path:
+    if task in ['train', 'predict'] and results_path and not do_grid_search:
         raise ValueError('The results path should not be set when task is '
-                         '{}.'.format(task))
+                         '{} and "grid_search" is set to False.'.format(task))
```
desilinguist: Does it make sense to fail out from `predict` too? That task doesn’t do anything with grid search and has no results.

mulhod: It looks like one can set grid search to True for `predict` as well as `train` and it will happen the same way in both cases. That is, if `train`, the model that results will use the best parameters. If `predict`, the model that gets trained and is used to predict on the test set will use the best parameters found during grid search. So, in the latter case, the results that are provided will be exactly the same as the former case. Is this incorrect?

desilinguist: Ah, of course. I forgot that `predict` will train a model. I thought it always loads an existing model. Hmm, actually what happens in this PR in the case where we are reusing an already trained model, e.g., by specifying `-k` at the command line?

mulhod: Good point. I'll take a look at that case. I'm not sure what will happen.

mulhod: In cases where a results path is set and grid search is on but then grid search is actually not done, results that are reported will basically have grid score/grid search CV results set to `null`, so the results will not be very informative (though seeing `null` values will tell the user grid search was not done, at least). There are two situations where this will happen:
1. When you're reusing a model. In this case, grid search has already been done (presumably) and there's nothing to do. The results that will be generated will overwrite the previous results file, however, which is not great if you wanted to use those results, but that's a pretty strange edge case. Perhaps a check could be introduced to not overwrite results if they exist and you're reusing a model that exists. That could be complicated.
2. When you're using an empty parameter grid, in which case there's no grid search to do. This might seem unlikely, but it's possible that the user specifies a learner whose default parameter grid is empty, like `LinearRegression`. In that case, the results will also be empty in terms of grid search-related attributes.

In both cases, I would argue that the behavior is expected. If you're using `-k`, you should know that grid search won't be done and setting grid search to true/specifying a results path is pointless. In the case where grid search won't be done due to there being no grid to search, the results related to the grid search will be accurate at the very least.

desilinguist: Thanks for investigating. I think this is fine but we will definitely need to update documentation to briefly mention some of these things. Please update the documentation issue to include this.

Issue #455, opened 2019-02-06
Make seaborn a direct dependency of SKLL


desilinguist:

desilinguist:
keep it open please

Issue #364, opened 2017-08-28
Move to Pandas to improve `Reader` and `FeatureSet` performance


jbiggsets:
Reading data into `FeatureSet` objects using the `Reader` classes seems to take a long time, compared with simply using `pandas` IO Tools. I did some quick benchmarking using a 200 MB TSV to test the following scenarios:

- Read TSV into DataFrame using `pd.read_csv()` function. *[9 seconds]*
- Read TSV into DataFrame using `pd.read_csv()` function, and then load `pd.DataFrame` to `FeatureSet` using `FeatureSet.from_data_frame()`. *[1 minute 17 seconds]*
- Load TSV into `FeatureSet` using `Reader.read()` method of `TSVReader` object. *[2 minutes 29 seconds]*

See attached for a copy of the Jupyter notebook I used for testing ([reader_speed_tests.zip](https://github.com/EducationalTestingService/skll/files/1257755/reader_speed_tests.zip)). For the last scenario, I also used `%prun` to get a sense of which function calls were taking the most time. Below are the top five function calls, ranked by total time (in seconds):

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)

 54600000  101.372    0.000  101.372    0.000 readers.py:685(safe_float)

   200002   22.954    0.000  141.354    0.001 readers.py:515(_sub_read)

        1   19.265   19.265   98.357   98.357 dict_vectorizer.py:133(fit_transform)

   200002    8.522    0.000   15.881    0.000 csv.py:108(__next__)

   200004    6.952    0.000    7.125    0.000 {built-in method builtins.next}
```

Recommendation: Move to `pandas` IO Tools in the next release, where operations like type conversion will be vectorized. This would obviously add a new dependency to `SKLL`, and there may be other considerations, but it seems like it would be worth it, if only for performance reasons.

desilinguist:
@jbiggsets thanks for the detailed report! I agree that it's time to start using `pandas` for reading tabular data in SKLL. It'd be weird to me if someone who wants to use `skll` would have a problem with `pandas` as a dependency since it's so pervasive now and so easily installable.

desilinguist:
Let's keep this open.

desilinguist:
Keep it open.

desilinguist:
Keep it open please.

desilinguist:
Keep it open please.

desilinguist:
Keep it open.

desilinguist:
[...]

PR #475, opened 2019-03-01
Add `pandas` and update `Readers` and `Writers`

jbiggsets:
This is a PR to address #364. 

**Summary**

This PR does the following

- Adds `pandas` as a direct dependency to `skll`, rather than an optional dependency.
- Updates several `skll.reader` and `skll.writer` classes to read/write in batch with `pandas`, rather than reading/writing line-by-line.
- Updates documentation according to the above changes.
- Updates tests and other files to accommodate the above changes.

**Additional Details**

Only certain `skll.reader` and `skll.writer` classes have been modified to read/write files using `pandas`. The reason for this is (1) `pandas` already offers a I/O functions for  TSV and CSV, and (2) ARFF and NDJ files (and DictList objects) can easily be manipulated to be read/written with `pandas`. Since this is not the case for MegaM and LibSVM, these files are still being read/written the old way, line-by-line.

I did not modify the `NDJWriter` to work with `pandas`, since  current approch (using `print` with `json.dumps`) seems to be equivalently fast to using `.to_json()`. Reader and writer classes that work with `pandas` have a  (private) `_use_pandas` attribute to indicate that the reading/writing should be performed in batch using `pandas`.

**Benchmarking**

I tested `skll`’s current `skll.Reader` and `skll.Writer` classes against the new implementations using `pandas` (with the C parser). The `pandas` implementation was consistently faster, but generally used more memory. The following tests were performed using a data set with 1 million rows and 100 feature columns (plus `y` and `id` columns). I also performed additional ad hoc testing with smaller data sets (1,000 and 10,000 records). In all of these tests, the new `skll.Reader` and `skll.Writer` classes ran faster than the old classes, while using on average 1.5 times the memory. 

The differences in speed were quite dramatic (especially with larger data sets) for all of the objects, excelt the `NDJReader`, where the new class (with `pandas`) performed only slightly faster than the old class. This is not a result of the new class being slow; rather, the old `NDJReader` class was by far the fastest of all the `skll.Reader` classes with larger data sets.

`skll.CSVReader` with `pandas`:

```
peak memory: 7843.98 MiB, increment: 6937.32 MiB
CPU times: user 4min 52s, sys: 39.6 s, total: 5min 31s
Wall time: 5min 31s
```

`skll.CSVReader` without `pandas`:

```
peak memory: 5438.30 MiB, increment: 5310.90 MiB
CPU times: user 36min 36s, sys: 8.93 s, total: 36min 45s
Wall time: 36min 47s
```

`skll.NDJReader` with `pandas`:

```
peak memory: 12005.62 MiB, increment: 11476.57 MiB
CPU times: user 5min 12s, sys: 41.6 s, total: 5min 54s
Wall time: 5min 56s
```

`skll.NDJReader` without `pandas`:

```
peak memory: 7171.23 MiB, increment: 6113.67 MiB
CPU times: user 7min 44s, sys: 10.23 s, total: 7min 54s
Wall time: 7min 56s
```

`skll.CSVWriter` with `pandas`:

```
peak memory: 9776.95 MiB, increment: 8253.72 MiB
CPU times: user 4min 20s, sys: 848 ms, total: 4min 21s
Wall time: 4min 22s
```

`skll.CSVWriter` without `pandas`:

```
peak memory: 1429.28 MiB, increment: 140.27 MiB
CPU times: user 57min 44s, sys: 10.23 s, total: 57min 54s
Wall time: 57min 56s
```

`skll.NDJWriter` with `pandas`:

```
peak memory: 3795.02 MiB, increment: 2247.75 MiB
CPU times: user 1min 37s, sys: 2.34 s, total: 1min 39s
Wall time: 1min 39s
```

`skll.NDJWriter` without `pandas`:

```
peak memory: 759.54 MiB, increment: 372.77 MiB
CPU times: user 51min 11s, sys: 1.43 s, total: 51min 12s
Wall time: 51min 12s
```
**Notes**
A separate PR will be submitted to address #455.


.travis.yml
```diff
@@ -25,9 +25,9 @@ before_install:
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then export PATH=/home/travis/miniconda2/bin:$PATH; else export PATH=/home/travis/miniconda3/bin:$PATH; fi
   - conda update --yes conda
 install:
-  - conda install --yes --channel defaults --channel conda-forge python=$TRAVIS_PYTHON_VERSION numpy scipy beautifulsoup4 six scikit-learn==0.20.1 joblib tabulate python-coveralls ruamel.yaml
+  - conda install --yes --channel defaults --channel conda-forge python=$TRAVIS_PYTHON_VERSION numpy scipy pandas beautifulsoup4 six scikit-learn==0.20.1 joblib tabulate python-coveralls ruamel.yaml
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes --channel defaults configparser mock; fi
-  - if [ ${WITH_PANDAS_AND_SEABORN} == "true" ]; then conda install --yes --channel defaults pandas seaborn; fi
+  - if [ ${WITH_PANDAS_AND_SEABORN} == "true" ]; then conda install --yes --channel defaults seaborn; fi
```
desilinguist: This should now just be called `$WITH_SEABORN` right?

skll/data/readers.py
```diff
@@ -95,6 +94,7 @@ def __init__(self, path_or_list, quiet=True, ids_to_floats=False,
         self.id_col = id_col
         self.class_map = class_map
         self._progress_msg = ''
+        self._use_pandas = False
```
desilinguist: Ah, so are we keeping the old way of reading in the files? What's the motivation for that?

.travis.yml
jbiggsets: Yeah, I was debating whether to change this. Since I have another PR coming that's removing `seaborn` (which means getting rid of this entirely), I figured I would just leave it for now.

Thoughts?

desilinguist: Sure, that makes sense.

skll/data/readers.py
jbiggsets: Good question, I should've explained this better above.

It makes sense to use `pandas` to read `TSV`, `CSV`, and `ARFF` files, since  either (1) `pandas` either already offers a reader for the file type (TSV, CSV), or (2) the data can easily be manipulated to be read with `pandas` (ARFF). However, that is not the case with `MegaM`, `NDJ`, `LibSVM`, and `DictList`.

The idea is that we'll keep the regular format (reading element-by-element) these file types that cannot be easily converted to be read with `pandas`. For `TSV`, `CSV`, and `ARFF`, we'll just read them directly into `pd.DataFrame` objects. The `_use_pandas` attribute just indicates whether the specific subclass uses a `pandas` reader, or continues doing things the old way.

jbiggsets: 1. I was debating whether to use `pandas` for some of the `writer` objects. I never benchmarked whether writing would be faster with `pandas`. I'll do some testing and see whether it's worth it.

2. Yes, absolutely, I'll to that, too.

3. Yup, I think this makes sense. I'll update.

desilinguist: Okay, so now the `use_pandas`  attribute makes sense. However, couldn't we handle NDJ/jsonlines files too using some combination of `apply()` and `pd.read_json()`?

Re: my engine comment, may be you were already thinking this but perhaps we should try and use the C engine and fall back to the Python engine if that fails.

skll/data/writers.py
```diff
@@ -221,145 +227,125 @@ def _write_line(self, id_, label_, feat_dict, output_file):
...
-        else:
+        # if there is no filtering, then just keep all the names
+        (column_names,
+         column_idxs) = self._get_column_names_and_indexes(feature_set,
+                                                           filter_features)
+
+        # create the data frame from the feature set;
+        # then, select only the columns that we want,
+        # and give the columns their correct names
+        df = pd.DataFrame(feature_set.features.todense())
```
desilinguist: Hmm, is there no way around converting things to `dense` and we are making a copy as well. Should we check if we are trading off a lot of memory again just like when reading?

skll/utilities/plot_learning_curves.py
```diff
@@ -68,7 +68,7 @@ def main(argv=None):
         makedirs(args.output_dir)
 
     # check that we have pandas and seaborn available
-    if not (_HAVE_PANDAS and _HAVE_SEABORN):
+    if not _HAVE_SEABORN:
         logging.error("Error: need pandas and seaborn to generate learning curve plots.")
```
desilinguist: Should we change the error message to remove "pandas and" since that's already installed by default now?

skll/data/readers.py
```diff
@@ -238,12 +241,127 @@ def feat_dict_generator():
...
+        """
+        if df.empty:
+            raise ValueError("No features found in possibly "
+                             "empty file '{}'.".format(self.path_or_list))
+
+        # if the id column exists,
+        # get them from the data frame and
+        # delete the column
+        if id_col is not None:
+            ids = df[id_col]
```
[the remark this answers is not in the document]
jbiggsets: Good catch. What I _should_ be doing here is checking `if id_col is not None and id_col in df`, and then forcing the `id_col` to `None` otherwise. I'll update.

skll/data/readers.py
```diff
@@ -337,50 +436,38 @@ def _sub_read(self, f):
...
-                try:
-                    curr_id = float(curr_id)
-                except ValueError:
-                    raise ValueError(('You set ids_to_floats to true, but' +
-                                      ' ID {} could not be converted to ' +
-                                      'float').format(curr_id))
-
-            yield curr_id, class_name, example
+        with open(f, 'r' if PY3 else 'rb') as buff:
+            lines = [json.loads(line.strip()) for line in buff
```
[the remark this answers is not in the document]
jbiggsets: OK, excellent. If we can assume there are not comments, this is much easier. I did some testing and it seems like it's faster, too. Still testing the memory trade-offs.

skll/data/writers.py
jbiggsets: I'm not sure of another way to convert a sparse array to a `DataFrame`, other than calling `toarray()` or `todense()`. Is `toarray()` more efficient? Or is there some alternative I'm not thinking of?

skll/utilities/plot_learning_curves.py
jbiggsets: Yup. I'll change this. I was being lazy about it because I'm removing `seaborn` anyway in a separate PR, so this whole error message is going away. :)

skll/data/readers.py
```diff
@@ -268,54 +386,31 @@ class DictListReader(Reader):
...
-            labels.append(class_name)
-            feat_dicts.append(example)
-            # Print out status
-            if example_num % 100 == 0:
-                self._print_progress(example_num)
-        # Convert lists to numpy arrays
-        ids = np.array(ids)
-        labels = np.array(labels)
-        features = self.vectorizer.fit_transform(feat_dicts)
+    def _sub_read(self, f):
```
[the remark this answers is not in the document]
mulhod: Hitching to this comment, it seems like the `f` of other `_sub_read` methods is also a bit confusing. For example, in `NDJReader`'s method, the documentation is:
```python
        """
        ...
        f : file buffer
            A file buffer for an NDJ file.
        ...
        """
```

However, `f` seems to be just a file path, judging by the way it's `open`ed: `with open(f, 'r' if PY3 else 'rb') as buff:`. I know it's not part of your PR, but I just wanted to mention it here.

skll/data/readers.py
```diff
@@ -765,68 +801,54 @@ def _sub_read(self, f):
...
+                self.relation = row_series.loc[0, 1]
 
-        # Process data as CSV file
-        return super(ARFFReader, self)._sub_read(chain([field_str], f))
+        df.columns = columns
+        return self._parse_dataframe(df, self.id_col, self.label_col)
 
 
-class TSVReader(DelimitedReader):
+class TSVReader(Reader):
```
mulhod: It might be nice if we could make `TSVReader` a sub-class of `CSVReader` with an instance attribute like `sep` set to `"\t"` (the default, defined in `CSVReader`, would be `","`). Then, `sep=self.sep` could replace e.g. `sep='\t'` in `pd.read_csv`. It would reduce the duplication of the two classes by a lot.

mulhod: The same thing could be done for `TSVWriter`/`CSVWriter`.

skll/data/readers.py
```diff
@@ -715,13 +717,34 @@ def _sub_read(self, f):
         features : list of dicts
             The features for the features set.
         """
-        try:
-            df = pd.read_csv(f, engine='c')
-        except CParserError:
-            df = pd.read_csv(f)
+        df = pd.read_csv(path, sep=self._sep, engine='c')
```
desilinguist: So the 'c' engine would never fail for this?

skll/data/readers.py
```diff
@@ -788,18 +811,15 @@ def split_with_quotes(s,
                   'quotechar': quote_char,
                   'escapechar': escape_char}
 
-        try:
-            df = pd.read_csv(StringIO(s), engine='c', **kwargs)
-        except CParserError:
-            df = pd.read_csv(StringIO(s), **kwargs)
+        df = pd.read_csv(StringIO(s), engine='c', **kwargs)
```
desilinguist: And same here?

skll/data/readers.py
```diff
@@ -788,18 +811,15 @@ def split_with_quotes(s,
...
         return df
 
     def _sub_read(self, f):
```
desilinguist: Should `f` be renamed to `path` then?

desilinguist:
Actually, @jbiggsets it would be nice if we could document how this line-by-line vs. pandas thing is effected through the use of `_sub_read()` and `_sub_read_rows()`. It doesn't need to be in the user-facing documentation - it can be in the code as well somewhere in the `Reader` class?

skll/data/readers.py
jbiggsets: I couldn't come up with a scenario where (1) the 'c' engine would fail, and (2) the 'python' engine wouldn't fail, _unless_ it's a situation where the user needs to pass additional arguments to `read_csv()` that aren't available with the 'c' parser.

If we want to allow users to pass additional arguments directly to the `pandas` reader, then I think this makes sense, but since we're not allowing them to do that, I don't see the advantage.

Let me know if you think I'm missing something.

desilinguist: Perhaps that would be a nice thing to have to allow for keyword argument to `pandas` in case we just can't think of a solution. Would it be too much work to add that?

jbiggsets: Well, it depends. :)

We could  just pass any additional keyword arguments to the `Reader` constructor onto `pd.read_csv()` (or whatever the `pandas` I/O function is). Alternatively, we could allow users to pass arguments to the constructor with a prefix, like `pandas_*` (e.g. `pandas_engine='python'`.), or something like that.

My only concern is that there are a lot of possible `pandas` arguments, so as long as we can be agnostic about what the `pandas` arguments are (and not have to test each one), this is easy.

Does that make sense? Thoughts?

desilinguist: I think we can follow the example of the `Learner` class where we have `model_kwargs` that are keyword arguments are just passed on to the underlying initializer. That should be simple enough right? So something like `pandas_kwargs`?

jbiggsets: Yup! Sounds good. I'll update.

desilinguist: Thanks for your continued patience :)

PR #476, opened 2019-03-15
Document the internal conversion of string labels to ints/floats and possible edge cases

mulhod:
I'm a bit unsure where this documentation should go. It would seem to be out of place and obtrusive almost anywhere except for in the `skll.readers` module itself. I tried to keep it brief and I hope that it can serve a purpose. Let me know if you think that it doesn't work here or should be written differently, etc.

### DIFF
--- a/doc/run_experiment.rst
+++ b/doc/run_experiment.rst
@@ -406,6 +406,26 @@ example, if you wanted to collapse the labels ``beagle`` and ``dachsund`` into a
 
 Any labels not included in the dictionary will be left untouched.
 
+One other use case for ``class_map`` is to deal with classification labels that
+would be converted to ``float``s improperly. All ``Reader`` sub-classes use the
+``safe_float`` function internally to read labels. This function tries to
+convert a single label first to ``int``, then to ``float``. If neither
+conversion is possible, the label remains a ``str``. Thus, care must be taken
+to ensure that labels do not get converted in unexpected ways. For example,
+consider the situation where there are classification labels that are a mixture
+of ``int``-converting ``float``-converting labels:
+
+.. code-block:: python
+    import numpy as np
+    from skll.data.readers import safe_float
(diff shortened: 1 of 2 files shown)


desilinguist:
This looks great. I think we could also add the same text as a note to the [`class_map` section](
https://skll.readthedocs.io/en/latest/run_experiment.html#class-map-optional) in the documentation.

mulhod:
Opened this up as a non-draft PR now.

doc/run_experiment.rst
```diff
@@ -406,6 +406,26 @@ example, if you wanted to collapse the labels ``beagle`` and ``dachsund`` into a
 
 Any labels not included in the dictionary will be left untouched.
 
+One other use case for ``class_map`` is to deal with classification labels that
+would be converted to ``float`` improperly. All ``Reader`` sub-classes use the
+``safe_float`` function internally to read labels. This function tries to
```
desilinguist: Please link `safe_float` to the API documentation. You should just be able to replace `safe_float` with :py:mod:`skll.data.readers.safe_float`.

doc/run_experiment.rst
```diff
@@ -406,6 +406,26 @@ example, if you wanted to collapse the labels ``beagle`` and ``dachsund`` into a
...
+convert a single label first to ``int``, then to ``float``. If neither
+conversion is possible, the label remains a ``str``. Thus, care must be taken
+to ensure that labels do not get converted in unexpected ways. For example,
+consider the situation where there are classification labels that are a mixture
+of ``int``-converting and ``float``-converting labels:
+
+.. code-block:: python
+    import numpy as np
```
desilinguist: You need an empty line between the `code-block` directive and the actual code - it causes an error on my machine otherwise.

desilinguist:
How the heck did coverage increase with this PR? 😛

PR #478, opened 2019-03-15
adding feature_scaling=both in boston example

bndgyawali:
This PR adds feature scaling to boston example.


examples/boston/cross_val.cfg
```diff
@@ -8,6 +8,7 @@ train_directory = train
 featuresets = [["example_boston_features"]]
 # there is only set of features to try with one feature file in it here.
 featureset_names = ["example_boston"]
+feature_scaling = both
```
desilinguist: Please add a comment about why this is something worth doing - to put the various features on the same scale.

examples/boston/evaluate.cfg
```diff
@@ -9,12 +9,12 @@ test_directory = test
 featuresets = [["example_boston_features"]]
 # there is only set of features to try with one feature file in it here.
 featureset_names = ["example_boston"]
+feature_scaling = both
```
desilinguist: Add same comment here too.

desilinguist:
Keep it open please.

Issue #461, opened 2019-02-13
Show warning when using feature hashing and there are multiple files


desilinguist:
Since the feature hashing applies to each specified feature file separately, it's worth showing a warning that clearly mentions this to the user since that might not be what the user is expecting.

desilinguist:
Addressed by #479.

PR #479, opened 2019-03-15
show warning when feature hashing is used for multiple featuresets

bndgyawali:
This PR addresses the issue #461.
It shows warning when feature hashing is used on multiple featuresets.
A unit test is added to confirm the behavior.


tests/test_output.py
```diff
@@ -95,6 +96,7 @@ def tearDown():
                 os.unlink(join(config_dir, cf))
 
         for output_file in (glob(join(output_dir, 'test_{}_*'.format(suffix))) +
+                            glob(join(output_dir, 'test_{}.log'.format(suffix))) +
```
mulhod: Is this actually removing all of the outputs or just the `.log` file? You probably need the asterisk here like the others.

bndgyawali: Actually, `test_{}_*` was not removing log files which would have format `test_{}.log`. So, I added this extra logic to remove log files.

desilinguist: But are there other files named `test_{}` (without the second underscore) that also need to be removed but aren't? Should you change `test_{}.log` to `test_{}.*`?

bndgyawali: I did not see any other files other than log. But, if there are, then they should be removed too. Let me check if any.

desilinguist:
Keep it open please

desilinguist:
@pep8speaks suggest diff

bndgyawali:
@desilinguist are you working on this branch? Else, I can confirm @mulhod's comment.

desilinguist:
Nope, I am not working on this branch. I just wanted to make sure the pep8 check bot was working. You can go ahead. Please address the pep8 suggestions as well as @mulhod's comments.

bndgyawali:
This is ready for review.

Issue #441, opened 2018-12-05
Capture scikit-learn warnings in log file


desilinguist:
Right now, any scikit-learn warnings (`FutureWarning`, `DeprecationWarning`, etc.) are printed to the console but do not get logged in the SKLL log files. There should be a way to capture those as well since these would be quite useful in seeing the changes that are coming to scikit-learn and hence SKLL.

desilinguist:
Addressed by #480

PR #480, opened 2019-03-21
Feature/log sklearn warnings

mulhod:
# Problem
When `sklearn` warnings are issued, they are printed to STDERR, but we want them to end up in the log.

# Solution
The `warnings.showwarning` method is what is used under the hood when one, say, issues a warning with, for example, `warnings.warn("here's a warning")`. That shows up as a `UserWarning`. One can set `warnings.showwarning` to a different method with the same signature and then, whenever a warning is issued, that new method will be run instead. In our case, that method would be overridden in the SKLL logger function (`logutils.get_skll_logger`). This new method will write the messages to the logger (including to an attached log file, if there is one) if the warning is from `sklearn` (others will be dealt with by calling the original `warnings.showwarning` method).

Another related issue (but not the main point of the issue related to this PR) is that we should close down the logger and make sure to close all file handlers attached to it. This is not so important for cases where we're running `run_configuration` once (especially via the command-line), but if we were to use `run_configuration` directly (or indirectly) in a Python script and we wanted to run it multiple times, then the issue of having open file handles, etc., becomes a problem. (I believe I have actually run into this problem before. It was puzzling that my jobs were failing after running a bunch of experiments due to an issue with too many open file handles. I couldn't figure out why so many file handles were being left open.) A later running of `run_configuration` would then potentially try to write to a log file from an earlier running. So, this PR adds in a utility for closing loggers and any attached file handles. That utility is used at the end of `run_configuration`, for example.


mulhod:
Just FYI, this build is expected to fail. There seems to be some problem specifically with the Python 2.7 test related to a test I added to check for warnings in the log while running `run_configuration`. It passes for Python 3.6 and actually it passes for 2.7 both on OSX and Linux on my local development platforms, just not on the Linux system where the build is run, unfortunately. But, anyway, I would at least like to open this up to discussion, if possible. I will mark it as WIP for now.

Also, one other note: The diff for this PR is scary, unfortunately. In `experiments.py` in particular, there are lots of "changes", but really all that's happening is that the bulk of two methods is wrapped in a `try`/`finally` and the contents are thus tab-shifted. Unfortunately, it's not that easy to tell what's happening.

desilinguist:
@mulhod perhaps we can do a walkthrough of this PR in the next sprint?

desilinguist:
Keep it open please.

skll/logutils.py
```diff
@@ -5,9 +5,14 @@
 :author: Nitin Madnani (nmadnani@ets.org)
 :organization: ETS
 """
-
+import re
```
bndgyawali: Can you put import statements in alphabetic order.

bndgyawali:
oops... is this ready for review? It still has `[WIP]`.

mulhod:
> oops... is this ready for review? It still has `[WIP]`.

Yes, sorry, it is ready now!

Please, everybody, review away!

tests/test_logutils.py
```diff
@@ -0,0 +1,141 @@
...
+        msg2 = "message 2"
+        logger.info(msg2)
+
+        with open(temp_file.name) as temp_file:
+            log_lines = temp_file.readlines()
+            assert log_lines[0].endswith("INFO - {}\n".format(msg1))
+            assert log_lines[1].endswith("INFO - {}\n".format(msg2))
+
+
+    def testGetSKLLLoggerWithWarning(self):
```
bndgyawali: can we make the name of the methods all lowercase with underscores rather than camel case.

tests/test_logutils.py
```diff
@@ -0,0 +1,141 @@
...
+            log_lines = temp_file.readlines()
+            assert log_lines[0].endswith("INFO - {}\n".format(msg1))
+            cv_sklearn_warning_re = \
+                re.compile(r"WARNING - [^\n]+site-packages/sklearn/model_selection/_split\.py"
+                           r":2053: FutureWarning:You should specify a value for 'cv' instead"
+                           r" of relying on the default value\. The default value will change"
+                           r" from 3 to 5 in version 0\.22\.")
+            assert cv_sklearn_warning_re.search("".join(log_lines[1:-1]))
+            n_estimators_sklearn_warning_re = \
+                re.compile(r"WARNING - [^\n]+site-packages/sklearn/ensemble/forest\.py:246: "
```
bndgyawali: I liked everything you did, but can this be some other warning rather than future warning. If this warning is removed or the functionality is absolute later on future, then we have to re-work to fix the test case too, isn't it?

tests/test_logutils.py
mulhod: Sure. I think that should be fine. I'm not sure why but it seems like the `unittest` module expects/uses camel case, though (e.g. `setUp`).

tests/test_logutils.py
mulhod: Yeah, that's a good point. Can you think of a case that raises a warning from `sklearn`? I have been trying to find a case to use and this was that worked out. But, yes, it will be a problem in the future.

bndgyawali: Let me search if I can find something.

mulhod:
Addresses #441.

skll/logutils.py
```diff
@@ -53,5 +58,38 @@ def get_skll_logger(name, filepath=None, log_level=logging.INFO):
             file_handler.setLevel(log_level)
             logger.addHandler(file_handler)
 
+    # Make method for sending sklearn-specific warnings to the
+    # logger that can be used to replace warnings.showwarning
+    def send_sklearn_warnings_to_log(message, category, filename, lineno,
```
desilinguist: There doesn't seem to be a real reason why this function is defined _inside_ `get_skll_logger()` other than just the fact that we want the function in scope? Perhaps a better solution would be as follows:

1. Rename `send_sklearn_warnings_to_log()` to `send_sklearn_warnings_to_logger()`, make it a top-level function in `logutils.py`, and add a first argument `logger`. 
2. Use the following assignment statement in `get_skll_logger()`: 

    ```python
    warnings.showwarning = partial(send_sklearn_warnings_to_log, logger)
    ```

This would make it so that there's a clear connection between the logger to be returned and the logger to which scikit-learn warnings should be sent. 

I just tested this and it seems to work. Thoughts @mulhod ?

skll/logutils.py
```diff
@@ -53,5 +58,38 @@ def get_skll_logger(name, filepath=None, log_level=logging.INFO):
...
+        else:
+            orig_showwarning(message, category, filename, lineno,
+                             file=file, line=line)
+    warnings.showwarning = send_sklearn_warnings_to_log
+
     # return the logger instance
     return logger
+
+
+def close_and_remove_logger_handlers(logger):
```
desilinguist: Thanks for adding this functions and catching the issue of unclosed file handlers.

tests/test_logutils.py
```diff
@@ -0,0 +1,136 @@
...
+    Do something that will trigger an ``sklearn`` warning.
+
+    This should trigger an ``UndefinedMetricWarning``.
+    """
+
+    roc_curve([1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+              [0, 1, 0, 1, 0, 1, 0, 1, 0, 1])
+
+
+class TestLogUtils(unittest.TestCase):
```
desilinguist: As much as I would prefer to switch SKLL tests to being class-based, that would require some serious refactoring and until then, I think it would be better if we stayed consistent with the way that the other tests are written and used module-level setup/teardown etc.

skll/logutils.py
mulhod: Seems to be working!

Issue #482, opened 2019-05-02
Default grid search parameters for MLPClassifier


SimengSun:
[https://github.com/EducationalTestingService/skll/blob/dcaa6d46311d8257f58b2124ebd7ae4866ccbab2/skll/learner.py#L121](url)

The default grid search parameters for MLPClassifier is now `[1e-4, 1e-3, 1e-3, 1e-1, 1]`,  maybe change one 1e-3 to 1e-2

PR #483, opened 2019-05-02
fixed default parameter duplication for HuberRegressor, MLPClassifier, MLPRegressor

Lguyogiro:
Addresses issue #482


Lguyogiro:
Not sure why coverage dropped here. None of the coverage changes seem related to the default parameter settings.

Lguyogiro:
I've added a test to ensure that the default parameter grids don't contain duplicate values, which was the issue that motivated this PR. 

This should be ready for review.

Lguyogiro:
@mulhod @bndgyawali Could either of you take a look at this PR? It is a small change, so should be pretty quick.

Issue #490, opened 2019-08-15
Update LogisticRegression documentation to talk about penalties


desilinguist:
We use the L2 regularization penalty by default but we should mention that both L1 and ElasticNet are also options if you choose the right solver.

PR #500, opened 2019-09-13
Update documentation for LogisticRegression penalty parameter

mulhod:
Discuss `penalty` parameter for `LogisticRegression`.

Addresses #490


mulhod:
#490

Issue #507, opened 2019-09-13
Update tests now that pandas and seaboard are direct dependencies


desilinguist:
We don't need to have selective tests anymore. We can just assume that pandas and seaborn are both _always_ available.

desilinguist:
Addressed by #508

Issue #492, opened 2019-08-21
`tabulate` missing in `conda_requirements.txt`


desilinguist:
Looks like when we switched to using `tabulate`, we added it to `requirements.txt` and the condo recipe but not to `conda_requirements.txt`.

desilinguist:
Addressed by #508

PR #508, opened 2019-09-14
Update various requirements files & remove selective testing

desilinguist:
- Add `tabulate` to all requirements files where it's missing (addresses #492).
- Add `seaborn` everywhere since it's now a direct dependency (addresses already closed issue #455).
- Update `.travis.yml` to remove selective testing based on `pandas` and `seaborn` since these are both now direct dependencies, and not optional (addresses #507).
- Switch to [codecov.io](https://codecov.io) since coveralls has been plagued with problems. Since codecov is not on `master`, it cannot compute the change from `master` but at least it computes the right coverage for the commit which is more than we can say for coveralls.


desilinguist:
coveralls is being stupid again.

PR #510, opened 2019-09-18
More Python 2.7 and Seaborn cleanups

desilinguist:
- There were still some files that were conditioning on `sys.version_info()` to do different things for python 2 vs python 3.
- There were still some checks based on `_HAVE_SEABORN` which aren't necessary since `seaborn` is now a direct dependency.


desilinguist:
@pep8speaks suggest diff

skll/experiments.py
```diff
@@ -15,9 +15,12 @@
...
 
+import matplotlib
+import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import ruamel.yaml as yaml
+import seaborn as sns
 
 from collections import defaultdict
 from io import open
```
bndgyawali: do we need this any more?

desilinguist: No, we don't. Good catch!

desilinguist: It's in a whole bunch of other files, I will fix those too.

bndgyawali: Yes. 
I saw these when grepping files under `skll`.
```
./config.py:16:from io import open
./utilities/summarize_results.py:14:from io import open
./experiments.py:26:from io import open
./data/readers.py:41:from io import open, StringIO
./data/writers.py:17:from io import open
```

Issue #536, opened 2019-10-01
`folds_file` does not get used for `evaluate` or `predict` tasks


desilinguist:
Right now, we only allow `grid_search_folds` to be read in from a specified `folds_file` if the `task` is `'train'` (see [here](https://github.com/educationalTestingService/skll/blob/master/skll/config.py#L854)). 

This means that the correct behavior is not exhibited for the `evaluate` or `predict` tasks which does not make sense since in terms of grid-search, these tasks are basically the same as the `train` task.

desilinguist:
Addressed by #538

PR #538, opened 2019-10-01
Allow `folds_file` to be used for `evaluate` and `predict` tasks.

desilinguist:
Addresses #536.

- Modify `config.py`  to allow `folds_file` to apply for not just `train` but also `evaluate` and `predict` since when it comes to grid search, the three tasks are equivalent.
- Update corresponding test in `test_input.py` to include the new tasks.
- Update documentation to mention the two new tasks.


tests/test_input.py
```diff
@@ -1269,6 +1269,24 @@ def test_cv_folds_and_grid_search_folds():
     # ('train', 'train/folds_file_test.csv', 7, None) ->  (None, fold_mapping)
     # ('train', 'train/folds_file_test.csv', 7, True) ->  (None, fold_mapping)
     # ('train', 'train/folds_file_test.csv', 7, False) ->  (None, fold_mapping)
+    # ('evaluate', None, None, None) ->  (None, 3)
```
ananyaganesh: What does (None, fold_mapping) represent?

desilinguist: As [this comment](https://github.com/EducationalTestingService/skll/blob/df6248695220b3662001812f1a3a66962450fe4a/tests/test_input.py#L1253) explains (probably not very clearly), that tuple represents the value of `cv_folds` and `grid_search_folds` that we expect to come out once `_parse_config_file()` has run.

skll/config.py
```diff
@@ -847,15 +847,15 @@ def _parse_config_file(config_path, log_level=logging.INFO):
...
-    #      contained in `grid_search_folds`.
+    #  (a) if the task is `train`/`evaluate`/`predict` and if an external
+    #      fold mapping is specified then use that mapping for grid search
+    #      instead of the value contained in `grid_search_folds`.
     #  (b) if the task is `cross_validate` and an external fold mapping is specified
-    #      then use that mapping for the outer CV loop. Depending on the value of
-    #      `use_folds_file_for_grid_search`, use the fold mapping for the inner
-    #       grid-search loop as well.
+    #      then use that mapping for the outer CV loop and for the inner grid-search
+    #      loop. However, if  `use_folds_file_for_grid_search` is `False, do not
```
mulhod: Nitpick: I think ``` `False``` should be changed to ``` `False` ```. I'm not really sure because this is inline comments. I actually thought values like this should have two pairs of ``` ` ```s, one at each end. We're also using this style (with single ``` ` ```) for other things like ``` `evaluate` ``` (in some cases, though we don't use anything).

desilinguist: yeah, we really need to be consistent here. I'll make the change.

Issue #537, opened 2019-10-01
Force `ids` to be strings for pandas-based readers


jbiggsets:
Currently in the `master` branch, if the user provides an ID column for the `CSVReader` or `TSVReader` classes, the IDs will automatically converted to the appropriate data type (`int` or `float`) if they are numeric, rather than being kept as strings. However, this is inconsistent with the _old_ behavior of these two classes, which  causes some downstream impacts if the reader wants to use a folds file. 

We should fix this for now by forcing the IDs to strings (unless `ids_to_floats` is passed), but we should also try to think of a better solution in the future. (I will file a separate issue to that effect.)

desilinguist:
Addressed by #539

PR #539, opened 2019-10-01
Fix pandas-based readers to read IDs as strings

desilinguist:
Addresses #537.

The pandas-based readers (`CSVReader` and `TSVReader`)  were automatically converting number-like IDs to the appropriate numeric datatype even when `ids_to_floats` was `False`. This is inconsistent with the other readers (e.g., `NDJReader`) that keeps those IDs as strings unless `ids_to_floats` is `True`. This PR enforces the same behavior for the pandas-based readers.

This should have been caught by the `skll_convert` tests in `test_utilities.py` but it was _not_ because those tests were all using explicit string IDs of the form `EXAMPLE_X`.  This PR also makes these tests more comprehensive by adding other types of IDs and also includes `ids_to_floats` as appropriate.

It also adds a little note about IDs in `readers.py` and add Jeremy's name as an author in `readers.py` and `writers.py`.


skll/data/readers.py
```diff
@@ -325,7 +332,7 @@ def _parse_dataframe(self, df, id_col, label_col, features=None):
         # delete the column; otherwise, just
         # set it to None
         if id_col is not None and id_col in df:
-            ids = df[id_col]
+            ids = df[id_col].astype(str)
```
bndgyawali: Can we not do this inside `_sub_read()` in `pd.read_csvmethod of `CSVReader` instead of doing it here. 

`df = pd.read_csv(file, sep=self._sep, engine=self._engine, dtype={self.id_col: str}, **self._pandas_kwargs)`

desilinguist: I don't think we can because at that point `self.id_col` may be `None`. `_parse_dataframe` checks for that internally.

bndgyawali: I think it works okay
```
In [13]: ! head -1 ./tests/train/folds_file_test.csv
id,cv_fold_num

In [14]: df = pd.read_csv("./tests/train/folds_file_test.csv", dtype={None: float})

In [15]: df.columns
Out[15]: Index(['id', 'cv_fold_num'], dtype='object')
```

desilinguist: But what's the advantage - there's no point in passing the argument `{None: float}` if you know it doesn't do anything?

jbiggsets: I also think it's smarter to do this inside the `_parse_dataframe` method, especially if we add other flat file readers (Excel, FWF) that rely on `pandas` in the future. We know that the IDs are always being forced to strings inside that method, regardless of what is passed to the underlying `pandas` reader.

Issue #541, opened 2019-10-02
Missing logger attribute in `FilteredLeaveOneGroupOut`


desilinguist:
In `FilteredLeaveOneGroupOut.split()`, we try to [log a warning](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L277-L278) if the feature set contains IDs that aren't in the folds dictionary. However, this statement will fail as this class does not have a `logger` attribute at all. We need to pass the logger from the specific learner down to this.

desilinguist:
Addressed by #543

PR #543, opened 2019-10-02
Add missing logger attribute to `FilteredLeaveOneGroupOut`

desilinguist:
Addresses #541 . 

- Passes the `Learner` logger down to `FilteredLeaveOneGroupOut` which means any warnings logged by the latter will get printed to the job log.
- Add a new test that triggers said warning which was not being tested before.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -211,7 +211,7 @@ def create_jsonlines_feature_files(path):
 
     # we only need to create the feature files if they
     # don't already exist under the given path
-    feature_files_to_create = [join(path, 'f{}.jsonlines'.format(i)) for i in range(5)]
+    feature_files_to_create = [join(path, 'f{}.jsonlines'.format(i)) for i in range(6)]
     if all([exists(ff) for ff in feature_files_to_create]):
         return
     else:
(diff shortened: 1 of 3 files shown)


desilinguist:
It was actually running a modified version of the example but adding the examples themselves as tests is a great idea! That way we don't have to manually test them for each release. Please file an issue?

tests/utils.py
```diff
@@ -241,9 +241,22 @@ def create_jsonlines_feature_files(path):
...
                 sub_features.append(x)
             fs = FeatureSet('ablation_cv', ids, features=sub_features, labels=labels)
+
             writer = NDJWriter(file_path, fs)
             writer.write()
 
+        # now write out the last file which is basically
+        # identical to the last featureset we wrote
+        # except that it has two extra instances
+        fs = FeatureSet('extra',
```
bndgyawali: I was having difficulty in understanding this. Please correct me if I am wrong
Though `create_jsonlines_feature_files` function is called from multiple locations, they are only using feature files between `f0` and `f4`. The only new test case `test_folds_file_with_fewer_ids_than_featureset` will use `f5` that has more instances than in the provided `folds_file` file.

desilinguist: `create_jsonlines_feature_files()` is created from multiple locations but it only does something if any of the files is missing from disk. But you are right in that `f5.jsonlines is _only_ used by this one test and not anywhere else.

Issue #544, opened 2019-10-03
Run example configs as part of tests


mulhod:
Lately we've encountered some issues where we don't realize that changes being introduced are breaking the examples (in generating the data or actually running the configs). Typically, we don't try each of the example configs every time we make a change, instead relying on the unit tests for that. So, we should create tests that simply generate the example data and run the configs. It might be a little challenging for one of the example data-sets since you have to retrieve it from the Kaggle site. (Update: Let's just download the required files and commit them to `skll`.)

bndgyawali:
Can we not add that dataset somewhere for free and does not require authentication and use it for test purposes?

desilinguist:
I agree with @bndgyawali. I feel like we can just commit the titanic dataset from Kaggle to our repo since it's just for our example anyway and plus that competition is so old - they are not changing the data in anyway.

mulhod:
Yes, that would be great. I will modify the issue.

jbiggsets:
Test the following:

## titanic
- [ ] `cross_validate.cfg`
- [ ] `evaluate_tuned.cfg`
- [ ] `evaluate.cfg`
- [ ] `learning_curve.cfg`
- [ ] `predict_train_only_tuned.cfg`
- [ ] `predict_train_only.cfg`
- [ ] `predict_train+dev_tuned.cfg`
- [ ] `predict_train+dev.cfg`
- [ ] `train.cfg`

## boston
- [ ] `cross_vali.cfg`
- [ ] `evaluate.cfg`

## iris
- [ ] `cross_vali.cfg`
- [ ] `evaluate.cfg`

desilinguist:
Addressed by #546

Issue #529, opened 2019-09-27
Make sure examples and tutorial work before 2.0 release


aoifecahill:
Currently the titanic example doesn't work (related to the fix for the readers). 

But we should also make sure that all other examples and the tutorial still work before the final release.

aoifecahill:
We should also add this step to the release documentation (so that we don't have to keep filing these issues for every release).

desilinguist:
Specifically, the titanic example won't work because `make_titanic_data.py` now behaves differently. In the old version -- before `pandas` was used as the backend for `CSVWriter` -- blank values were incorrectly backfilled to be 0s (see [this line](https://github.com/EducationalTestingService/skll/blob/v1.5.3/skll/data/writers.py#L319)). However, with `pandas`, the correct behavior happens, and blank values remain blank. 

So, we basically need to use the `pandas_kwargs` keyword argument in the new `CSVReader` class and filter out those rows which have any empty values. Does that make sense @jbiggsets?

desilinguist:
Hmm, actually it looks like there's no keyword argument to pandas' `read_csv` that will filter out rows with NaNs _while_ reading. So, we can just use `FeatureSet.filter()` instead.

jbiggsets:
Test the following:

## titanic
- [ ] `cross_validate.cfg`
- [ ] `evaluate_tuned.cfg`
- [ ] `evaluate.cfg`
- [ ] `learning_curve.cfg`
- [ ] `predict_train_only_tuned.cfg`
- [ ] `predict_train_only.cfg`
- [ ] `predict_train+dev_tuned.cfg`
- [ ] `predict_train+dev.cfg`
- [ ] `train.cfg`

## boston
- [ ] `cross_vali.cfg`
- [ ] `evaluate.cfg`

## iris
- [ ] `cross_vali.cfg`
- [ ] `evaluate.cfg`

desilinguist:
@jbiggsets see also #544

mulhod:
@jbiggsets Can you add the same comment over there, actually?

desilinguist:
Addressed by #546.

PR #546, opened 2019-10-04
[WIP] Update examples, and create tests

jbiggsets:
This PR is intended to address #529 and #544. It is a work in progress.


tests/test_examples.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Run all of the configuration files for the boston example
+    """
+    for config_path in glob(join(_new_boston_dir, 'boston', '*.cfg')):
+        run_configuration(config_path, quiet=True)
+
+
+def test_iris_configs():
+    """
+    Run all of the configuration files for the boston example
```
jbiggsets: Will change this to `"iris"` in the docstring.

tests/test_examples.py
```diff
@@ -0,0 +1,105 @@
...
+_old_titanic_dir = join(_examples_dir, 'titanic')
+_old_boston_dir = join(_examples_dir, 'boston')
+_old_iris_dir = join(_examples_dir, 'iris')
+
+_new_titanic_dir = join(_my_dir, 'other', 'titanic')
+_new_boston_dir = join(_my_dir, 'other', 'boston')
+_new_iris_dir = join(_my_dir, 'other', 'iris')
+
+
+def run_function_in_custom_dir(func, old_dir, new_dir):
```
mulhod: This seems fine for now. Although it might be nice to fix these scripts so that they don't require being run from a particular directory so that this wouldn't be necessary.

desilinguist: Could we just use `subprocess.run()` here to run the script which allows specifying the working directly explicitly?

tests/test_examples.py
```diff
@@ -0,0 +1,105 @@
...
+    rmtree(_new_boston_dir)
+    rmtree(_new_iris_dir)
+
+
+def test_titanic_configs():
+    """
+    Run all of the configuration files for the titanic example
+    """
+    for config_path in glob(join(_new_titanic_dir, 'titanic', '*.cfg')):
+        run_configuration(config_path, quiet=True)
```
desilinguist: So, we are just running the examples? Wouldn't a better test be to load the results JSON file (the path is returned by `run_configuration`) and check some expected values? For example, what if the tests run in a future update but return different values? Wouldn't we want to know?

jbiggsets: Yeah, I debated this, too. The issue was just to make sure that the examples don't _break_ after we've made updates, not that the values haven't changed (which I assume they would in many cases).

But I can add some checks for expected values, maybe with a lower degree of precision?

desilinguist: I think it would be useful if the values changed significantly - that should indicate that something went wrong on our end or the sklearn folks have changed something major that we should be paying attention to. 

I agree that lower precision (may be 2 decimal places) should be sufficient. I don't think the values should change more than that with a routine upgrade right?

tests/test_examples.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Run the configuration, and then check the JSON results
+    against expected JSON files
+    """
+
+    # run this experiment, get the `results_json_path`
+    results_json_path = run_configuration(config_path, local=True, quiet=True)[0]
+
+    # if the results path exists, check the output
+    if exists(results_json_path):
```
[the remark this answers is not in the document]
jbiggsets: No, I don't think so. For example, the Titanic Predict example don't produce `results.json`.

tests/test_examples.py
```diff
@@ -0,0 +1,136 @@
...
+            # we obviously want to skip any keys that we aren't expecting
+            if key in results_exp_obj:
+
+                actual = results_obj[key]
+                expected = results_exp_obj[key]
+
+                # if this is a float, then we check with less precision (2 decimals);
+                # otherwise, we check to make sure things are matching exactly
+                if isinstance(expected, float):
+                    assert_almost_equal(actual, expected, places=2)
```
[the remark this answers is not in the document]
jbiggsets: That sounds OK to me. 👍

tests/test_examples.py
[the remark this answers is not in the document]
desilinguist: Hmm, I guess if something went wrong with the `run_configuration()` command, it would raise an error anyway?

Issue #548, opened 2019-10-08
`neg_log_loss` is incorrect when specified as an additional metric


desilinguist:
When [computing additional output metrics](https://github.com/educationalTestingService/skll/blob/master/skll/learner.py#L1801-L1822), we are not paying close enough attention to whether the metric needs probabilities. We only use the probability values when it comes to computing correlations but not when it comes to say `neg_log_loss` - where we are incorrectly [passing in the labels](https://github.com/educationalTestingService/skll/blob/master/skll/learner.py#L1818-L1820).

desilinguist:
The other probabilistic metrics (`average_precision` and `roc_auc`) were also not being computed correctly.

desilinguist:
Addressed by #551

Issue #545, opened 2019-10-03
Correlation metrics not working as expected for binary classification


desilinguist:
Back when we wrote an early version of SKLL, we wanted a way to use correlation metrics as tuning objectives for binary classification problems in such a way that we could use the probability of the positive class as the input to the correlation function rather than just the label itself. This should yield a more informative value for the objective. 

The way we we achieved this was as follows: before starting the grid search, if we [see](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L1642-L1643) that we had a correlation metric and a classifier, we [monkey-patch](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L1644-L1645) the estimator being passed to the `GridSearchCV` instance - replacing its `predict()` method with the custom [`_predict_binary()`](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L400) function. When this function is called, it checks whether the classification being effected is binary and if so, it uses the probability of the positive class as the prediction and. If it's not binary, it outputs the most likely label as the prediction. 

However, this monkey-patching is no longer working as this [notebook](https://github.com/EducationalTestingService/skll/files/3688038/correlation-test.pdf) shows. Basically, even for the binary case, the most likely label is being used when computing pearson. 

To fix this, we should no longer need this monkey-patching. We can simply do the following: 

- If we see that the `probability` option is enabled and that the objective is, say, `'pearson'`, we dynamically update `SCORERS['pearson']` to be `make_scorer(pearson, needs_proba=True)` rather than `make_scorer(pearson)` which is the default value. This will make it such that our  [`pearson()` function in `metrics.py`](https://github.com/EducationalTestingService/skll/blob/master/skll/metrics.py#L192) will receives probabilities as the estimator predictions if the `probability` configuration option is enabled and the most likely labels as the predictions otherwise.

- In the `pearson()` function itself, we first check that the array of predictions is 1-dimensional (which will be the case for both binary classification since that's how `needs_proba` works), then we just compute the pearson correlation of the predictions with the true labels. However, if the array of predictions has dimension > 1 (which will be the case for multi-class classification), we just pick the most likely label and use that to compute the pearson correlation with the true label. The latter is because it doesn't make sense to use the probabilities in the multi-class case. 

With these modifications, all use cases with pearson as the objective function will work seamlessly for the user:

#### Case 1: binary classification, `probability` option enabled. 

[...]

PR #551, opened 2019-10-11
Overhaul computation and documentation of tuning objectives and metrics

desilinguist:
We have several bugs in how SKLL computes various classification metrics and which metrics it allows under which classification scenarios:
- Correlation metrics for probabilistic binary classification are not being computed correctly (#545). 
- Probabilistic metrics (`neg_log_loss`, `average_precision`, and `roc_auc`) were being computed with labels when they should just require `probability` to be `True` (#548). 
- We were not checking whether QWK was being used for non-contiguous integer labels (which is the only case where it makes sense).
- We are not checking that correlations are only being computed for non-string labels.
- The documentation was not clear about which metrics could be used in what context.

This PR includes the following:

**Major Changes**
- Modifies the correlation metric functions defined in `metrics.py` to work with both probability arrays and arrays of labels.
- Modifies `Learner.train()` such that it dynamically modifies the correlation grid objective to use the version that works with probabilities when the goal is probabilistic binary classification.
- Adds a new function `_contiguous_ints_or_floats()` that checks whether the sorted, unique set of labels being used is contiguous. Also adds a corresponding test in `test_classification.py`.
- Adds two new functions `_get_acceptable_regression_metrics()` and `_get_acceptable_classification_metrics()` that return the list of acceptable metrics for different scenarios. 
- Adds explicit checks for tuning objectives and output metrics using the functions in the last two bullets. 
- Adds explicit checks for probabilistic metrics in `config.py` and corresponding test in `test_input.py`.
- Adds comprehensive tests in `test_classification.py` and `test_regression.py` for invalid metrics/objectives.
- Adds comprehensive tests in `test_classification.py` and `test_regression.py` to ensure that SKLL is computing metric/objective values identical to what scikit-learn (or scipy) would compute.
- Overhauls documentation for tuning objectives and adds an explicit note about how classification metrics are computed in SKLL.

**Minor Changes**
- Changes documentation author to `Educational Testing Service`.
- Modifies `make_titanic_data.py` script to work properly with `drop_blanks` and adds documentation about how to do this properly for CSV/TSV format files.
- Makes minor PEP-8 fixes here and there.
- Moves some functions around for readability. 
- Removes some things from top-level SKLL imports.


skll/learner.py
```diff
@@ -282,6 +250,44 @@ def split(self, X, y, groups):
...
+        ints_or_int_like_floats = np.all(np.mod(numbers, 1) == 0)
+
+        # next check that the successive differences between
+        # the numbers are all 1, i.e., they are nuermicontiguous
+        contiguous = np.all(np.diff(numbers) == 1)
+
+    except TypeError:
+        raise TypeError('Input should only contain numbers.')
+
+    # we need both comditions to be true
```
mulhod: Nitpick: typo: comditions -> conditions

skll/learner.py
```diff
@@ -282,6 +250,44 @@ def split(self, X, y, groups):
...
+    return ints_or_int_like_floats and contiguous
```
mulhod: Nitpick: It may be an unnecessary edge case to address, but if the input is an empty list, this function will return `True`.

```
>>> _contiguous_ints_or_floats([])
True
```

skll/learner.py
```diff
@@ -282,6 +250,44 @@ def split(self, X, y, groups):
             yield train_index, test_index
 
 
+def _contiguous_ints_or_floats(numbers):
+    """
+    Summary
```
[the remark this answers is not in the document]
desilinguist: Yes, it should! Nice catch!

skll/learner.py
```diff
@@ -397,33 +403,78 @@ def _train_and_score(learner,
...
+    # are acceptable
+    if issubclass(label_type, (np.object_, str)):
+        pass
+
+    # CASE 2: labels are integers or floats; the way
+    # it works in SKLL, it's guaranteed that
+    # class indices will be sorted in the same order
+    # as the class labels therefore, ranking metrics
+    # such as various correlations should work fine.
+    elif issubclass(label_type, (int, np.int64, float, np.float64)):
```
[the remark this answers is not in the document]
desilinguist: Hmm, I guess technically it could be. Are there really 32-bit machines out there in the wild that people might use to run machine learning experiments? I guess we never know. I'll include these too.

skll/learner.py
```diff
@@ -397,33 +403,78 @@ def _train_and_score(learner,
...
-        res = self.predict_normal(X)
-    return res
+    # this is a classifier so the acceptable objective
+    # functions definitely include those metrics that
+    # are specifically for classification and also
+    # the unweighted kappa metrics
+    acceptable_metrics = _CLASSIFICATION_ONLY_METRICS | _UNWEIGHTED_KAPPA_METRICS
+
+    # now let us consider which other metrics may also
+    # be acceptable depending one whether the labels
```
mulhod: typo: "one" -> "on"

skll/metrics.py
```diff
@@ -164,14 +204,30 @@ def kendall_tau(y_true, y_pred):
...
+    if y_pred.ndim > 1:
+        labels = np.argmax(y_pred, axis=1)
+        ret_score = kendalltau(y_true, labels)[0]
+    # 1-dimensional -> probabilities/labels -> use as is
+    else:
+        ret_score = kendalltau(y_true, y_pred)[0]
     return ret_score if not np.isnan(ret_score) else 0.0
 
 
 def spearman(y_true, y_pred):
```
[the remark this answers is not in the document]
desilinguist: Hmm, I didn't really think about that. I am not sure how easy or readable it would be since we these are essentially being passed to `make_scorer()` but I will try and take a look.

skll/learner.py
```diff
@@ -397,33 +403,78 @@ def _train_and_score(learner,
...
+
+    # CASE 2: labels are integers or floats; the way
+    # it works in SKLL, it's guaranteed that
+    # class indices will be sorted in the same order
+    # as the class labels therefore, ranking metrics
+    # such as various correlations should work fine.
+    elif issubclass(label_type, (int, np.int64, float, np.float64)):
+        acceptable_metrics.update(_CORRELATION_METRICS)
+
+        # CASE 3; labels are numerically contiguous integers
```
mulhod: Nitpick: the semicolon should probably be a colon

skll/learner.py
```diff
@@ -1665,6 +1707,13 @@ def train(self, examples, param_grid=None, grid_search_folds=3,
             grid_score = 0.0
             grid_cv_results = None
 
+        # restore the original of the grid objective if we
+        # had futzed with the it to handle correlation
```
mulhod: Nitpick: probably need to delete "the"

tests/test_classification.py
```diff
@@ -779,3 +857,620 @@ def test_sampling_for_multinomialNB():
...
+            sklearn_fold_test_probs = sklearn_learner.predict_proba(X_fold_test)
+
+            # now let's proceed on a metric by metric basis
+            # and note that we are again using the class
+            # indices rather than the class labels themselves
+
+            # 1. Correlation metrics are only computed for integer
+            #    or float labels; they use probability values if
+            #    available only in the binary case.
+            if metric_name == 'pearson':
```
mulhod: Nitpick: There is some shared code between the blocks related to the conditions for Pearson, Spearman, and Kendall-Tau. They could probably be collapsed and a `metric_func`-type variable could be used to loop over the 3 possible metrics.

tests/test_classification.py
```diff
@@ -779,3 +857,620 @@ def test_sampling_for_multinomialNB():
...
+    sklearn_test_probs = sklearn_learner.predict_proba(X_test)
+
+    # now let's proceed on a metric by metric basis
+    # and note that we are again using the class
+    # indices rather than the class labels themselves
+
+    # 1. Correlation metrics are only computed for integer
+    #    or float labels; they use probability values if
+    #    available only in the binary case.
+    if metric_name == 'pearson':
```
mulhod: Same comment as above.

skll/metrics.py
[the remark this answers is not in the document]
desilinguist: It wasn't that hard actually and it does cut down on duplication! Thanks for the suggestion!

tests/test_classification.py
desilinguist: @jbiggsets had the same suggestion! I just pushed a commit that does this. Thanks!

tests/test_classification.py
desilinguist: I don't want to do the refactoring here because I don't want to run as little custom SKLL code as possible in a test.

desilinguist: I don't want to do any refactoring here as I want to run as little SKLL code during tests as possible.

skll/learner.py
desilinguist: Good catch! I can make this return an exception.

skll/metrics.py
```diff
@@ -164,14 +204,30 @@ def kendall_tau(y_true, y_pred):
...
     """
     Calculate Spearman's rank correlation coefficient between ``y_true`` and
-    ``y_pred``.
+    ``y_pred``. ``y_pred`` can be multi-dimensional. If ``y_pred`` is
+    1-dimensional, it may either contain probabilities, most-likely
+    classification labels, or regressor predictions. In that case, we
+    simply return the pearson correlation between ``y_true`` and ``y_pred``.
```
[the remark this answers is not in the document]
desilinguist: This is already outdated! I refactored all 3 functions into a single one :)

tests/test_classification.py
mulhod: I'm not sure I get what you mean. I just meant that you could have something like the following:
```python
if metric_name in ['pearson', 'spearman', 'kendaulltau']:
    if metric_name == 'pearson':
        metric_func = pearsonr
    elif metric_name == 'spearman':
        metric_func = spearmanr
    else:
        metric_func = kendalltau
    if issubclass(label_type, (np.int64, np.float64)):
        if len(label_array) == 2 and use_probabilities:
            sklearn_metric_value = metric_func(y_test,
                                            sklearn_test_probs[:, 1])[0]
        else:
            sklearn_metric_value = metric_func(y_test,
                                            sklearn_test_labels)[0]
```

It's a little clunky now that I see it, but I think it does reduce duplication a bit.

desilinguist: Oh I see what you mean! That's a good idea 👍.

examples/make_titanic_example_data.py
```diff
@@ -92,6 +99,13 @@ def main():
     test_fs = Reader.for_path('test.csv',
```
mulhod: It looks like `test.csv` does not have a `Survived` column:
```
➜  titanic git:(streamline-objectives) ✗ head test.csv
PassengerId,Pclass,Name,Sex,Age,SibSp,Parch,Ticket,Fare,Cabin,Embarked
892,3,"Kelly, Mr. James",male,34.5,0,0,330911,7.8292,,Q
```

This is the error I get:
```
(/Users/mmulholland/Documents/skll_dev/skll_env) ➜  titanic git:(streamline-objectives) ✗ ../../skll_env/bin/python ../make_titanic_example_data.py
2019-10-14 13:05:28,426 - skll.data.readers - INFO - Rows/lines with any blank values will be dropped.
Loading train.csv...           done
Writing titanic/train/vitals.csv...done
Writing titanic/train/socioeconomic.csv...done
Writing titanic/train/family.csv...done
Writing titanic/train/misc.csv...done
Writing titanic/train+dev/vitals.csv...done
Writing titanic/train+dev/socioeconomic.csv...done
Writing titanic/train+dev/family.csv...done
Writing titanic/train+dev/misc.csv...done
Writing titanic/dev/vitals.csv...done
Writing titanic/dev/socioeconomic.csv...done
Writing titanic/dev/family.csv...done
Writing titanic/dev/misc.csv...done
Traceback (most recent call last):
  File "../make_titanic_example_data.py", line 124, in <module>
    main()
  File "../make_titanic_example_data.py", line 109, in main
    sparse=False).read()
  File "/Users/mmulholland/Documents/skll_dev/skll/data/readers.py", line 427, in read
    ids, labels, features = self._sub_read(self.path_or_list)
  File "/Users/mmulholland/Documents/skll_dev/skll/data/readers.py", line 844, in _sub_read
    df = pd.read_csv(file, sep=self._sep, engine=self._engine, **self._pandas_kwargs)
  File "/Users/mmulholland/Documents/skll_dev/skll_env/lib/python3.7/site-packages/pandas/io/parsers.py", line 685, in parser_f
    return _read(filepath_or_buffer, kwds)
  File "/Users/mmulholland/Documents/skll_dev/skll_env/lib/python3.7/site-packages/pandas/io/parsers.py", line 457, in _read
    parser = TextFileReader(fp_or_buf, **kwds)
  File "/Users/mmulholland/Documents/skll_dev/skll_env/lib/python3.7/site-packages/pandas/io/parsers.py", line 895, in __init__
    self._make_engine(self.engine)
  File "/Users/mmulholland/Documents/skll_dev/skll_env/lib/python3.7/site-packages/pandas/io/parsers.py", line 1135, in _make_engine
    self._engine = CParserWrapper(self.f, **self.options)
  File "/Users/mmulholland/Documents/skll_dev/skll_env/lib/python3.7/site-packages/pandas/io/parsers.py", line 1959, in __init__
    _validate_usecols_names(usecols, self.orig_names)
  File "/Users/mmulholland/Documents/skll_dev/skll_env/lib/python3.7/site-packages/pandas/io/parsers.py", line 1255, in _validate_usecols_names
    "columns expected but not found: {missing}".format(missing=missing)
ValueError: Usecols do not match columns, columns expected but not found: ['Survived']
```

desilinguist: Good catch! Will fix.

skll/learner.py
```diff
@@ -1465,56 +1533,55 @@ def train(self, examples, param_grid=None, grid_search_folds=3,
...
+                                                             label_type.__name__))
+
+            # If we're using a correlation metric for doing binary
+            # classification and probability is set to true, we assume
+            # that the user actually wants the `_with_probabilities`
+            # version of the metric
+            if (grid_objective in _CORRELATION_METRICS and
+                    estimator_type == 'classifier' and
+                    self.probability):
+                self.logger.info('You specified "{}" as the objective with '
```
mulhod: The value that is supposed to be inserted in `{}` is never actually supplied (via `.format(...)`.

mulhod: This is a nice message and I confirmed that I could get it when expected (after making the abovementioned change):
```
2019-10-14 14:06:53,666 - Iris_CV_example_iris_MultinomialNB - INFO - You specified "spearman" as the objective with "probability" set to "true". If this is a binary classification task with integer labels, the probabilies for the positive class will be used to compute the correlation.
```

However, this statement does get logged a lot each time `train` is called. If you're doing cross-validation, it happens as many times as there are folds. I wonder if something can be done about that.

Also, due to the way it's being called right before the training is done, it's possible that somebody could write a config file that has multiple objectives and it goes through almost all the experiments and then fails on the last (say, because you are using `spearman`). I don't see an easy way of sort moving this check "up". But it would be nice if possible to fail early.

I think these two things are related in some way: if you could fail early, you wouldn't need to log it for each fold. You would just log it at that earlier point once.

skll/learner.py
```diff
@@ -1465,56 +1533,55 @@ def train(self, examples, param_grid=None, grid_search_folds=3,
...
+                                 '"probability" set to "true". If this is '
+                                 'a binary classification task with integer '
+                                 'labels, the probabilies for the positive '
```
mulhod: Nitpick: type in "probabilies"

skll/learner.py
desilinguist: It would be nice to have these but I don't see a way of doing this without adding a lot more complexity and I don't think the annoyance of failing late or seeing multiple log messages warrants that. 

From another point of view, I'd say that this is just the price you pay for using the more convenient option of the config file. For example, we also read in the training file for every single sub-experiment if you have multiple learners and objectives. If these things annoy someone, they can just use the API directly.

desilinguist: It would be nice to have these but I don't see a way of doing this without adding a lot more complexity and I don't think the annoyance of failing late or seeing multiple log messages warrants that.

From another point of view, I'd say that this is just the price you pay for using the more convenient option of the config file. For example, we also read in the training file for every single sub-experiment if you have multiple learners and objectives. If these things annoy someone, they can just use the API directly.

Issue #562, opened 2019-10-16
Allow specifying of `pos_label_str` when using the `Predictor()` class


aoifecahill:
For binary classification experiments, there is a bit of a disjunction when you specify the positive label string in the configuration. In this case the labels do not get sorted alphabetically. 

Sometimes it is necessary to specify `pos_label_str` in the config file so that the model will get tuned correctly. However in the `Predictor` class in `generate_predictions.py` we are being forced to specify a `pos_label_index`. Unless you explicitly look inside the model `label_dict`, you won't know for sure which class is in which index (since by specifying `pos_label_str`, the classes do not get sorted alphabetically). 

It would probably make more sense to have the `Predictor` class accept the same `pos_label_str` and then figure out the index based on the `label_dict`. 

The documentation for generate predictions assumes that the labels have been sorted. This should be clarified too.

desilinguist:
This and #550  should be handled together.

desilinguist:
I think this is a broader issue:

1. We don't need the `Predictor` class. We should just be able to use `Learner.from_file()` and then `Learner.predict()`.
2. We shouldn't even need to provide `pos_label_str`.  We can just either use what’s stored in the model file or, if it's not there, infer it the way the learner would have (alphabetical sorting).
3. This script should just operate in 3 modes: predict most likely labels (binary/multiclass/regression), predict label with custom threshold (binary only), and predict probabilities for all classes (binary/multiclass).

desilinguist:
Addressed by #566.

Issue #550, opened 2019-10-09
Write explicit tests for `pos_label_str`


desilinguist:
We don't have any tests for `pos_label_str` and string labels. It would be good to add some tests to make sure that things are working as expected.

desilinguist:
Addressed by #566.

PR #566, opened 2019-10-19
Major fixes for `pos_label_str`

desilinguist:
Addresses #550 and #562.

- Modify `Learner._create_label_dict()` such that `pos_label_str` is ignored (and forcibly set to `None`) if specified for multi-class classification.
- Further modify `Learner._create_label_dict()` to use `if self.pos_label_str is None` instead of `if self.pos_label_str` since the latter might lead to buggy behavior if the integer `0` is passed as the value of `pos_label_str`.
- Fix bug in how the value of `pos_label_str` is read in via the configuration file. Since `pos_label_str` is supposed to represent the positive class label, we should be applying `safe_float()` to it just like we do to feature labels when they are read in. If we don't do this, then `pos_label_str` will not behave correctly when a value of `0` is passed in via the config file since the config file would read it as the string `'0'` but the labels in the data are read in as the number `0`.
- Also apply `safe_float()` to value of `pos_label_str` when passed in through the `Learner` constructor just to be on the safe side.
- Update the description for `pos_label_str` in the `Learner` class docstring.
- Overhaul code for `generate_predictions`
    - Get rid of `Predictor` class and just use `Learner.predict()`.
    - Get rid of `pos_label_index` and just use the `pos_label_str` from the learner instance, if available, otherwise infer it just like the learner would.
    - Get rid of old options and use simpler modes of operation.
- Update documentation for `generate_predictions`.
- Add a new test for label index ordering that interacts with `pos_label_str`.


desilinguist:
Code coverage decrease in `generate_predictions.py` is expected as we are significantly shortening it.

skll/utilities/generate_predictions.py
```diff
@@ -187,21 +92,47 @@ def main(argv=None):
...
+    estimator_type = learner.model_type._estimator_type
+
+    # if we are using a binary classification model, get the positive
+    # class label string if from the `pos_label_str` attribute, and if
+    # that is `None`, get it from the learner's label dictionary; also
+    # get the string denoting the negative label which is just the other
+    # label in the list
+    if estimator_type == 'classifier':
+        if len(learner.label_list) == 2:
+            if learner.pos_label_str:
```
jbiggsets: Could `pos_label_str` still be 0 here? If so, shouldn't this also be `if learner.pos_label_str is not None`?

desilinguist: Good catch!

desilinguist:
I am adding more tests so that should help with coverage.

skll/learner.py
```diff
@@ -869,9 +870,13 @@ class Learner(object):
         initializer for the specified model.
         Defaults to ``None``.
     pos_label_str : str, optional
```
mulhod: Is it true that this will necessarily be a `str` here? I'm a little confused because it seems like it could be a float or int at this point too (in addition to `None`).

tests/utils.py
```diff
@@ -158,7 +158,8 @@ def fill_in_config_options(config_template_path,
                             'suffix'],
                   'Tuning': ['grid_search', 'objective',
                              'use_folds_file_for_grid_search', 'grid_search_folds',
-                             'param_grids', 'objectives', 'duplicate_option'],
+                             'pos_label_str', 'param_grids', 'objectives',
```
mulhod: Why is `pos_label_str` in the tuning section? I always thought it was a bit of bad fit. I usually assume it's in the input section.

skll/learner.py
desilinguist: Yes, it could be anything but it's easier to say `str` because even if the user puts `"1"` or `"1.0"` or anything, `safe_float()` will handle it just fine. So, I didn't want to make the docstring overly complicated.

desilinguist:
@mulhod if you can share the data and the config file, that'd be quite useful.

skll/learner.py
mulhod: I'm mainly concerned with the documentation. If we were to use a type annotation, this would probably need to be changed to be `Optional[Union[str, float, int]]`.

desilinguist: Sure, we can do that when it comes time to do that.

skll/learner.py
```diff
@@ -910,7 +915,7 @@ def __init__(self, model_type, probability=False, pipeline=False,
         self.scaler = None
         self.label_dict = None
         self.label_list = None
-        self.pos_label_str = pos_label_str
+        self.pos_label_str = safe_float(pos_label_str) if pos_label_str is not None else pos_label_str
```
bndgyawali: And, continuing the comment from `config.py`, is the `pos_label_str` will never be None but will be `None` string coming here, so do we need `safe_float` here?

skll/config.py
```diff
@@ -606,7 +607,7 @@ def _parse_config_file(config_path, log_level=logging.INFO):
                                                     "sampler_parameters"))
     fixed_sampler_parameters = yaml.safe_load(fixed_sampler_parameters)
     param_grid_list = yaml.safe_load(_fix_json(config.get("Tuning", "param_grids")))
-    pos_label_str = config.get("Tuning", "pos_label_str")
+    pos_label_str = safe_float(config.get("Tuning", "pos_label_str"))
```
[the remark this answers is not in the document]
desilinguist: No, because the [default value](https://github.com/EducationalTestingService/skll/blob/master/skll/config.py#L85) in config.py for `pos_label_str` is the empty string `''`. So, the value will still be the empty string which basically gets happily ignored when doing the sorting in `Learner._create_label_dict()`. But you are right in that we should make the default values of `pos_label_str` be `None` both via the API and the configuration file. I will make a PR for this.

skll/learner.py
desilinguist: It won't be `None` string so this does not apply.

Issue #564, opened 2019-10-17
Not all metrics being computed


aoifecahill:
If you specify multiple tuning objectives in the `Tuning` section of the config and multiple metrics in the `Output` section of the config, the output metrics are only computed for a metric if it does not appear in the `Tuning` section. 

This does not make sense in the case where you specify multiple objective metrics in the `Tuning` section, since you only want to suppress the calculation of the specific tuning objective in a single experiment, not all possible tuning objectives across all experiments. 

**More concretely**
In Tuning I specify the following: `objectives=[f1_score_weighted, accuracy, precision, average_precision, roc_auc]`
and then in output, I specify the following: `metrics=[f1_score_weighted, accuracy, precision, recall, roc_auc, average_precision]`
But say the `f1_score_weighted` metric will only be computed on the test data in the case where I am tuning on f1_score_weighted. And it will get reported under the grid_objective column in the summary file 
I would like it to also get computed (as an additional metric) when I am tuning on `accuracy`, `precision`, `average_precision`,  and `roc_auc`. 

I think this means the check for what metrics to exclude from calculation needs to be moved from [`config.py`](https://github.com/EducationalTestingService/skll/blob/master/skll/config.py#L840) to `learner.py`.

desilinguist:
Addressed by #567

PR #567, opened 2019-10-20
Fix bug in metric computation when there is overlap with grid objective

desilinguist:
This PR addresses #564. 

- Moves the overlap checking from the configuration level (in `config.py`) to the experiment level (in `learner.py`). 
- Adds another check for code added in the earlier metric computation PR: the message about probabilities being used to compute the metric is now _not_ shown when computing the objective since doing so might send conflicting messages to the user given the overlap message.
- Remove old test for overlap checking from `test_input.py`
- Add new test for overlap checking to `test_classification.py`.


tests/test_classification.py
```diff
@@ -199,8 +204,8 @@ def check_label_index_order_with_pos_label_str(use_api=True):
         # if we are using the API, create a learner instance directly
         # otherwise run a configuration file to train a learner
         if use_api:
-            clf = Learner('LogisticRegression', pos_label_str=pos_label_str)
-            clf.train(train_fs, grid_search=False)
+        clf = Learner('LogisticRegression', pos_label_str=pos_label_str)
```
bndgyawali: should not this be indented?

tests/test_classification.py
```diff
@@ -311,7 +316,7 @@ def check_binary_predictions_for_pos_label_str(label_list,
         clf_skll = Learner('LogisticRegression',
                            pos_label_str=pos_label_str,
                            probability=probability)
-        clf_skll.train(train_fs, grid_search=False)
+    clf_skll.train(train_fs, grid_search=False)
```
bndgyawali: this too.

tests/test_classification.py
```diff
@@ -356,11 +361,11 @@ def test_binary_predictions_for_pos_label_str():
          use_pos_label_str,
          probability,
          use_api) in product([['B', 'C'],
-                              [1, 2],
-                              ['FRAG', 'NONE']],
+                                  [1, 2],
```
bndgyawali: this too.

tests/test_classification.py
desilinguist: Weird, not sure how that got unindented.

tests/test_classification.py
```diff
@@ -1668,3 +1673,69 @@ def test_metric_values_for_classification():
...
+                                         values_to_fill_dict,
+                                         'metrics_objective_overlap',
+                                         good_probability_option=True)
+
+    # run this configuration file and get the output JSON
+    # files for each experiment
+    results_json_paths = run_configuration(config_path, local=True, quiet=True)
+
+    # make a dummy objective if we do not have any for the
+    # purposes if iteration
```
aoifecahill:
```suggestion
    # purposes of iteration
```

Issue #573, opened 2019-10-21
Cannot load saved `MultinomialNB` models


desilinguist:
We cannot load any saved `MultinomialNB` model files at the moment because `Learner.from_file()` tries to [look inside](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L1094) the `__dict__` attribute of the learner's `scaler` attribute and because explicitly disallow feature scaling for `MultinomialNB` (it doesn't support negative values), `scaler` is just `None`.

This basically means no one has ever tried to load a `MultinomialNB` model ever :)

desilinguist:
Addressed by #574

PR #568, opened 2019-10-21
Major update to SKLL documentation & other fixes

desilinguist:
💥 ** This PR must be reviewed after #566 and #567  ** 💥 

This PR addresses #459, #460, and #447 (which was recently reopened).

- Move *Output files* into its own major section right under *Running experiments* and add explicit, linkable sub-sections for each type of output file that SKLL can produce. Write documentation for each type of output file.
- Link these types of files from the corresponding Output section option.
- Rename *Creating configuration files* ➡️ *Configuration file fields* and  *Feature file formats* ➡️ *Feature files*. 
- Move the general workflow of a SKLL experiment from the end of the tutorial to the beginning of *Running experiments*.
- Sort all optional fields alphabetically in the *Configuration file fields*  section.
- Add `--no_labels` argument to `skll_convert` documentation.
- Add feature hashing warning about multiple files to documentation.
- Add note about change to default value of grid search config file option.
- Add note about feature hashing in `print_model_weights` documentation and to the docstring for `Learner.model_params()`.
- Update the learner fixed parameters in the documentation to match the code.
- Sort requirements in README.
- Update `CONTRIBUTING.md` to talk explicitly about conda environment and several other additions/tweaks.
- Add note about `pandas` being the backend for CSV/TSV files.
- Add `--sort_by_labels` to `print_model_weights` documentation.
- Remove `-c defaults` from installation command.
- Add note about `MultinomialNB` not support feature sampling.
- Document `save_cv_models` option in documentation. 
- Update year in license.
- Replace lambda with function as a PEP8 fix.
- Remove unneeded imports
- Clean up example data and `test_examples.py`
  -  For some reason we were relying on `titanic/titanic` which must have been a holdover from when we used to download the data.
  - Move the data files to under `examples/titanic` directly.
  - Clean up `test_examples`.
  - Update tutorial  to refer to already included titanic data files and not mention downloading from Kaggle anymore.
- Fix links in documentation.

PR #574, opened 2019-10-21
Fix loading issue with `MultinomialNB` model files.

desilinguist:
💥 **This PR should be reviewed after #567, #568, #571** 💥 

This PR Addresses #573.


aoifecahill:
Shouldn't we add a test for this to make sure we don't break it in future?

desilinguist:
Yes, we should! Adding one now 👍

desilinguist:
@aoifecahill test has been added.

PR #578, opened 2019-10-22
Removed create_label_dict attribute from Learner

### LINKED ISSUE #565: Remove `create_label_dict` argument from `Learner.train()`
We are using this public keyword argument to keep track of some internal state. It really shouldn't be visible to the user because there should be no reason why a user would want to set this to `False`. We should find another way to keep track of whether the label dictionary has already been created once.

Sando1:
Fixes #565 
Removed the attribute create_label_dict. 
Instead works on the self.label_dict parameter to achieve the same result.


desilinguist:
@Sando1 are you still interested in working on this? If so, can you please merge `master` in and resolve the conflicts? Also, it would be nice to have a test for this.

desilinguist:
This PR is not sufficient.

PR #580, opened 2019-10-24
Preparing for SKLL v2.0 Release 

desilinguist:
(**Note**: Due to the issues discovered on Windows which necessitated the integration of Azure Pipelines for regular testing, this release PR will be longer than usual. Thanks for your patience!)

- Code changes to get tests to run and pass on Windows
  - Explicitly tell SKLL readers and writers to read and write files in UTF-8. 
  - Add `newline=''` whenever writing out plain text files (summary files, folds file, etc.) to prevent Windows from adding an extra carriage return in CSV/TSV files.
  - Modify `test_input.py` and `test_logutils.py` to deal with "file is already open" errors on Windows.
  - Explicitly close `matplotlib` figure instances to prevent increased memory usage.
  - Modify the custom `NumpyTypeEncoder` JSON encoder to also include `np.int32` since those can be produced on Windows.
  - We cannot figure out why the test `test_send_warnings_to_log()` does not pass on Windows even though capturing sklearn warnings on Windows, so we are just skipping it for now. I have filed an issue to investigate it in the future.

- Add Windows testing
  - Add `azure-pipelines.yml` to configure the Azure Pipelines build.
  - Add `DistributeTests.ps1` powershell script to split up tests to run in parallel the same way they are done for Travis CI.
  - Remove Appveyor integration.

- Update `conda` recipe
  - Update version number.
  - Use a more efficient command for the build script that does not involve installing depenencies again as part of the build process. 
  - Sort build and run requirements
  - Add license file and update license short name.
  - Update README for recipe.

- Update documentation
  - Add installation section to `README.rst` and point to the installation section on ReadTheDocs.
  - Move requirements section up in `README.rst`.
  - Update conda badge in `README.rst`.
  - Update installation command in `getting_started.rst`.
  - Update release process documentation to add package building and testing commands.

- Sort entries in `requirements.txt`
- Update version number in `version.py`.
- Update `setup.py` and `MANIFEST.in` to exclude tests, examples, and documentation from the PyPI package.

#### Please run the following tests before approving this PR:

- Go through the [ReadTheDocs documentation](https://skll.readthedocs.io/en/release-2.0/) that I built specifically for this branch.
- Test the conda package by either installing it in your environment (`conda install -c conda-forge -c ets skll`), or creating an environment with it (`conda create -n skllenv -c conda-forge -c ets skll`). These commands will use python 3.7. You can also add `python=3.6` before `skll` to test with python 3.6.
- To test the TestPyPI package using `pip`, use `pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple skll`.
- Please test on your own Macs, Linux servers, as well as Windows, if possible.


doc/internal/release.rst
```diff
@@ -15,28 +15,34 @@ This document is only meant for the project administrators, not users and develo
...
 
-5. Upload each of the packages to anaconda.org using ``anaconda upload <package tarball>``.
+6. Upload the source distribution to TestPyPI  using ``twine upload --repository testpypi dist/*``. You will need to have the ``twine`` package installed and set up your ``$HOME/.pypirc`` correctly. See details `here <https://packaging.python.org/guides/using-testpypi/>`__.
 
-6. Upload source package to PyPI using ``python setup.py sdist upload``.
+7. Test the conda package by creating a new environment on different platforms with this package installed and then running SKLL examples or tests from a SKLL working copy. If the package works, then move on to the next step. If it doesn't, figure out why and rebuild and re-upload the package.
 
-7. Draft a release on GitHub.
+8. Test the TestPyPI package by installing it as follows: ``pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple skll`` and then running SKLL examples or tests from a SKLL working copy. If the package works, then move on to the next step. If it doesn't, figure out why and rebuild and re-upload the package.
```
bndgyawali: nitpick
`pip install` command is going out of the page, can we make it to multiple lines. i looked in chrome and safari and both have similar view. 
Screen Shot 2019-10-24 at 9 44 42 AM

requirements.txt
```diff
@@ -1,9 +1,9 @@
-scikit-learn==0.21.3
```
bndgyawali: I see `requirements.txt` and `conda_requirements.txt` are exactly same. Only difference is `conda_requirements.txt` is not sorted. Can we not have one of them.

desilinguist: Yeah, RSMTool consolidated they two requirements file (via work that @jbiggsets did) which is very cool. However, I would like to see how that pans out before making the same change here in SKLL. Let’s file an issue for this and plan for it for v2.1. 

For now, I will also sort `conda_requirements.txt`.

Issue #579, opened 2019-10-23
Investigate why the logging test does not pass on Windows


desilinguist:
Currently, we are skipping the test `test_send_warnings_to_log()` in `test_output.py` on Windows because we cannot figure out why it fails even though capturing sklearn warnings works in general.

desilinguist:
Addressed by #586

PR #586, opened 2019-10-28
Fix Windows log/warnings issue

mulhod:
Addresses #579.

Because of an issue with the way we were computing correlation, the log/warnings behavior was reacting differently on Windows vs \*nix. So, this PR just makes the test use a different objective that wouldn't issue the same warning. Also, an issue will be made to deal with log/warnings behavior when `grid_search_jobs` is greater than 1 in unit tests.


mulhod:
One more review, please? @bndgyawali @jbiggsets @aoifecahill @ananyaganesh @chaomenghsuan

PR #605, opened 2020-05-04
Remove `create_label_dict` keyword argument from `Learner.train()`

### LINKED ISSUE #565: Remove `create_label_dict` argument from `Learner.train()`

desilinguist:
This PR closes #565. 

- It also modifies the `Learner._create_label_dict()` argument so that it internally keeps track of whether `label_dict` has already been set.
- No additional tests are necessary since the cross-validation and learning curve tests already represent scenarios where this function is called twice, once in the top-level `cross_validate()`/`learning_curve()` methods and once again when the underlying models are trained.

### DIFF
--- a/skll/learner/__init__.py
+++ b/skll/learner/__init__.py
@@ -690,7 +690,7 @@ class Learner(object):
             The examples to use for training.
         """
 
-        # we don't need to do this if we have alerady done it
+        # we don't need to do this if we have already done it
         # or for regression models, so simply return.
         if (self.label_dict is not None or
                 self.model_type._estimator_type == 'regressor'):
--- a/skll/learner/__init__.py
+++ b/skll/learner/__init__.py
@@ -689,8 +689,11 @@ class Learner(object):
         examples : skll.FeatureSet
             The examples to use for training.
         """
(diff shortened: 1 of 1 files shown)


aoifecahill:
Looks good to me. What about the drop in code coverage (1 line?). Is that just noise?

desilinguist:
> Looks good to me. What about the drop in code coverage (1 line?). Is that just noise?

Yup, we can ignore that.

skll/learner/__init__.py
```diff
@@ -689,8 +689,11 @@ def _create_label_dict(self, examples):
         examples : skll.FeatureSet
             The examples to use for training.
         """
-        # We don't need to do this for regression models, so return.
-        if self.model_type._estimator_type == 'regressor':
+
+        # we don't need to do this if we have alerady done it
```
mulhod, line 693:
```suggestion
        # we don't need to do this if we have already done it
```

PR #612, opened 2020-05-22
Add support for custom metrics

### LINKED ISSUE #606: Add support for custom metrics
A good long term solution is to add support for custom metrics just like we have for custom learners. This means we won't need to keep adding new metrics and users can use any metric they want as long as it takes two vectors as input: the labels and the predictions. 

The machinery for implementing this would be as follows:

1. We create a new (private) mutable set called `_CUSTOM_METRICS` in `skll/metrics.py`.

2. We also add a new function in the same file called `register_custom_metric()` that takes in a path to a Python file containing a function that takes two arraylikes as arguments and returns a number. Whenever this function is called, it imports the metric function into the namespace, adds a new entry to the `SCORERS` dictionary in `skll/__init__.py`, and adds the metric name to the `_CUSTOM_METRICS` set.

3. We add a new configuration field called `custom_metric_path` that accepts a path to a Python file, just like `custom_learner_path`.

4. In `experiments.py`, we validate this path and call `register_custom_metric()` to "register" any  custom metrics specified under `objectives` or `metrics` to `SCORERS`.

5. We modify `get_acceptable_classificiation_metrics()`  and `get_acceptable_regression_metrics()` to return not just the built-in acceptable metrics but also any custom metrics. It's the user's responsibility to make sure that they are specifying the right custom metric for the right problem type.

6. For the API, users can call `register_custom_metric()` directly for each metric and then use the metric names in `learner.train()` etc. directly.

### LINKED ISSUE #609: Add jaccard metrics as acceptable classification metrics
scikit-learn offers [multiple variants of `jaccard` metrics](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.jaccard_score.html#sklearn.metrics.jaccard_score) that we should allow to be used in SKLL.

### LINKED ISSUE #610: Add other variants of precision and recall to SKLL
scikit-learn offers the following variants of precision and recall that would be useful to have in SKLL as well since we already offer the same variants for F1 and F0.5:

- `precision_macro`
- `precision_micro`
- `precision_weighted`
- `recall_macro`
- `recall_micro`
- `recall_weighted`

desilinguist:
This PR closes #606, closes #609, and closes #610.

**Custom Metrics**:

1. SKLL can now use arbitrary custom metric functions for both tuning and evaluation.
2. The details of how this works is described in detail in #606. The only tricky bit not described there is how the dynamic import works. Essentially, we take the Python file specified by the user and import it as a sub-module of `skll.metrics` and then tell `SCORERS` that the function that the name points to `skll.metrics.<filename>.<function>`.
3. Add a new file `test_custom_metrics.py` to test the custom metric functionality. 
4. Add a new section to the documentation dedicated to this functionality called "Using custom metrics". 

**Add new metrics**:

1. Added all variants of the `jaccard_score` metric from scikit-learn since those can be quite useful.
2. Added the non-binary variants of `precision` and `recall` to be consistent with `f1_score` and `f0.5_score`.
3. Added all new metrics to the documentation.


skll/metrics.py
```diff
@@ -215,6 +227,95 @@ def f1_score_least_frequent(y_true, y_pred):
...
+    """
+    Import, load, and register the custom metric function from the given path.
+
+    Parameters
+    ----------
+    custom_metric_path : str
+        The path to a custom metric.
+    custom_metric_name : str
+        The name of the custom metric function to load.
+        This function must take only two array-like
```
aoifecahill, line 240: Doesn't the documentation in `custom_metrics.rst` say that there are also three optional keyword arguments?

desilinguist: For the arbitrary custom metric function that _you_ are writing, not for this function :)

aoifecahill: ah, of course

skll/config/utils.py
```diff
@@ -170,16 +173,39 @@ def _parse_and_validate_metrics(metrics, option_name, logger=None):
...
+            (custom_file is not None and custom_file != '')):
+
+        # try to register each potential metric
+        for custom_metric in possible_custom_metrics:
+            try:
+                register_custom_metric(custom_file, custom_metric)
+
+            # do not raise an error if the registration fails since
+            # we will raise an exception for all invalid metrics later
+            except AttributeError:
```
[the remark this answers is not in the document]
desilinguist: Good catch! 

1. The `AttributeError` is raised as a result of the `getattr()` call in `register_custom_metric()`, i.e., when you are trying to import a function that does not exist in the file. You are right that it should be listed in the doctoring even though it's note explicitly raised in the code.

2. I will include the other exceptions here. I am a fan of explicitly listing out the exceptions rather than just using the base exception which is generally not recommended.

3. I also like your idea of updating `invalid_metrics`  instead of using `pass`.

desilinguist:
Okay, I have modified the implementation such that custom metrics are now properly serialized when using `gridmap`. However, in order to support `gridmap`, we had to unfortunately lose the functionality wherein any invalid metrics in the configuration file could be identified at configuration parsing time. This check is now deferred until right before the job is submitted. However, this shouldn't be too bad. This deferral needed to happen because the registration of any potential custom metrics now happens inside `_classify_featureset()` and we cannot declare any metrics as invalid before that. This meant removing a couple of config parsing tests. 

@mulhod can you please re-run your `gridmap` tests and any other tests you can think of. Thanks! @bndgyawali if you have time to review this too, it'd be great!

PR #617, opened 2020-06-02
Update scikit-learn to v0.23.1

### LINKED ISSUE #613: Update scikit-learn to v0.23.1
See full release notes for v0.23.1 [here](https://scikit-learn.org/stable/whats_new/v0.23.html#version-0-23-1).

desilinguist:
This PR closes #613. 

- Update `scikit-learn` version in all requirements files and in the conda recipe.
- Update SKLL version number to 2.5.0 in `version.py` and modify `Learner.from_file()` to raise an error when loading models trained using previous versions of SKLL.
- Remove accommodations in `Learner.from_file()` for really old versions of SKLL. 
- Update the `_get_param_names()` classmethod in the `rescaled` class decorator to use `inspect.signature()` rather than `inspect.getargspec{)` to match what's done in scikit-learn.
- Update the backwards compatibility test to use this SKLL version instead of 0.9.17 since we are now breaking backwards compatibility with all previous versions of SKLL. This test is pretty much useless for this version but it's in there for future releases. 
- Update the data for some tests and retrain models where needed.


desilinguist:
We can ignore the codecov “failures” as there are no other tests to add and they are just the byproduct of code being removed/refactored.

PR #619, opened 2020-06-05
Simplify parameter grids

### LINKED ISSUE #618: Simplify parameter grids to be dictionaries instead of lists of dictionaries
Right now, we always use a list with a single dictionary to represent the default parameter grid for a learner. However, [`GridSearchCV`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html) can take a plain dictionary as well. Given that we almost never search multiple grids, switching to a plain dictionary will make things a lot simpler when implementing voting and stacking learners.

desilinguist:
This PR closes #618. 

- Change the default parameter grids to be plain dictionaries instead of lists with a single dictionary. 
- Update docstrings and documentation to make it more consistent and compatible with this change. 
- Update tests.


desilinguist:
Almost all of our tests in `test_classification.py` and `test_regression.py` are with empty parameter grids since we do not specify an explicit parameter grid either via the API or via the config file. So that's already covered.

Issue #488, opened 2019-08-15
Add VotingClassifier and VotingRegressor ensembles.


desilinguist:
These would be great for baseline ensembles.

desilinguist:
This is actually a bit more complicated because it requires passing in the base estimators whose predictions will be voted on. This requires a bit more thinking,.

PR #622, opened 2020-06-19
Refactor code in preparation for voting learners

### LINKED ISSUE #621: Buggy interaction in `Learner.predict()` when used via API
There's a bug in `Learner.predict()` that only surfaces via the API for probabilistic learners. 

```python
from skll import Learner
from skll.data import NDJReader

l = Learner('SGDClassifier', probability=True)
train_fs = NDJReader.for_path('examples/iris/train/example_iris_features.jsonlines').read() 
test_fs = NDJReader.for_path('examples/iris/test/example_iris_features.jsonlines').read()
l.train(train_fs, grid_search=False)
```

Now, if I want to get the most likely labels in memory but write out the class probabilities to disk, I can actually do it the way the API is written:

```python
labels = learner.predict(test_fs, class_labels=True, prediction_prefix='blah')
```

Doing so actually raises the following error because we try to write out the probabilities even though we end up only computing labels in the code (due to `class_labels` being `True`):

```python
/work/skll/skll/learner/__init__.py in predict(self, examples, prediction_prefix, append, class_labels)
   1448                     for example_id, class_probs in zip(example_ids, yhat):
   1449                         print('\t'.join([str(example_id)] +
-> 1450                                         [str(x) for x in class_probs]),
   1451                               file=predictionfh)
   1452                 else:

TypeError: 'numpy.int64' object is not iterable
```

We should either explicitly disallow this case (it's a little weird to want labels in memory but probabilities on disk) or fix the code so that this work correctly. 

The latter would mean:

1. If `learner.probability` is `True`, always write out probabilities else write out the class labels.
2. If `class_labels` is `True`, return class labels else return the class indices.

Thoughts?

desilinguist:
As I was working on implementing a `VotingLearner` class (for #488), I realized that this implementation would end up sharing a lot of the code with methods for the `Learner` class. One way to handle this is to make `VotingLearner` inherit from `Learner` but that's not really right because, in fact, `VotingLearner` uses multiple underlying `Learner` instances. So, the solution I came up with is to refactor as much of this shared code into functions in `learner.utils` that can the be used by both `Learner` instances as well as `VotingLearner` instances (and, hopefully, `StackingLearner` instances too, when we get to those). 

Specifically, this PR refactors:

- code that computes evaluation metrics from `Learner.evaluate()` into `learner.utils.compute_evaluation_metrics()`. 
- code that writes predictions to files from `Learner.predict()` into `learner.utils.write_predictions()`. 
- code that merges any unseen test set labels with training labels from `Learner.evaluate()` into `learner.utils.add_unseen_labels()`.
- code that computes the number of folds based on the number of examples for each label from `Learner._compute_num_folds_from_example_counts()` into `learner.utils.compute_num_folds_from_example_counts()`.
- code that computes various types of predictions (raw, class labels, class indices, probabilities) from `Learner.predict()` into `learner.utils.get_predictions()`.
- code that sets up the cross-validation fold iterator from `Learner.cross_validate()` into `learner.utils.setup_cv_fold_iterator()`. 
- code that sets up the cross-validation split iterator from `Learner.learning_curve()` into `learner.utils.setup_cv_split_iterator()`.

In addition, This PR closes #621  and adds a test to make sure that the predictions being returned and written out via `Learner.predict()` match expectations. As part of this fix, the default value of the `class_labels` keyword argument for `Learner.predict()` is now set to `True` instead of `False` since it doesn't make sense to return class indices by default.
💥 **This will be a an API-breaking change since to get class probabilities as outputs, `class_labels` will explicitly need to be set to `False` .** 💥

Finally, this PR improves many docstrings, replaces single quotes with double quotes in some places, and replaces the old-style `format` strings with new-style ones in some of the code that was touched.


desilinguist:
Not sure why it doesn't show here but the Travis builds are actually complete.

desilinguist:
Actually, this is not done - just found another simplification I can make. Please don't review it yet.

desilinguist:
@bndgyawali @mulhod @aoifecahill this is now ready for review. The minor code coverage decrease is expected due to the refactoring.

skll/learner/utils.py
```diff
@@ -171,6 +181,260 @@ def _get_support_mask(self):
         return mask
 
 
+def add_unseen_labels(train_label_dict, test_label_list):
+    """
+    Merge test set labels that not seen in the training data with seen ones.
```
mulhod, line 186:
```suggestion
    Merge test set labels that are not seen in the training data with seen ones.
```

skll/learner/utils.py
```diff
@@ -171,6 +181,260 @@ def _get_support_mask(self):
...
+    ----------
+    train_label_dict : dict
+        Dictionary mapping training set class labels to class indices.
+    test_label_list : list
+        List containing labels in the test set.
+
+    Returns
+    -------
+    train_and_test_label_dict : dict
+        Dictionary mapping merged lables from both the training and test set
```
mulhod, line 198:
```suggestion
        Dictionary mapping merged labels from both the training and test sets
```

skll/learner/utils.py
```diff
@@ -171,6 +181,260 @@ def _get_support_mask(self):
...
+    """
+    # get the list of labels that were in the training set
+    train_label_list = list(train_label_dict.keys())
+
+    # identify any unseen labels in the test set
+    unseen_test_label_list = [label for label in test_label_list
+                              if label not in train_label_list]
+
+    # create a new dictionary for these unseen labels with label indices
+    # for them starting _after_  those for the training set labels
```
mulhod, line 209:
```suggestion
    # for them starting _after_ those for the training set labels
```

skll/learner/utils.py
```diff
@@ -171,6 +181,260 @@ def _get_support_mask(self):
...
+    metrics : list of str
+        List of metrics to compute.
+    labels : array-like
+        True labels to be used for computing the metrics.
+    predictions : array-like
+        The predictions to be used for computing the metrics.
+    model_type : str
+        One of "classifier" or "regressor".
+    label_dict : dict, optional
+        Dictionary mapping classes labels to indices for classification.
```
mulhod, line 242:
```suggestion
        Dictionary mapping class labels to indices for classification.
```

skll/learner/utils.py
```diff
@@ -171,6 +181,260 @@ def _get_support_mask(self):
...
+
+    return res
+
+
+def compute_num_folds_from_example_counts(cv_folds,
+                                          labels,
+                                          model_type,
+                                          logger=None):
+    """
+    Calculate the number of folds we should use for cross validation, based
```
mulhod, line 389:
```suggestion
    Calculate the number of folds we should use for cross-validation, based
```

skll/learner/utils.py
```diff
@@ -171,6 +181,260 @@ def _get_support_mask(self):
...
+    -------
+    cv_folds : int
+        The number of folds to use, based on the number of examples
+        for each label.
+
+    Raises
+    ------
+    ValueError
+        If ``cv_folds`` is not an integer or if the training set has
+        less than or equal to one label(s) for classification.
```
mulhod, line 415:
```suggestion
        fewer than 2 examples associated with a label (for classification).
```

skll/utils/commandline/generate_predictions.py
```diff
@@ -142,9 +142,9 @@ def main(argv=None):
         # make sure each file extension is one we can process
         input_extension = os.path.splitext(input_file)[1].lower()
         if input_extension not in EXT_TO_READER:
-            logger.error(('Input file must be in either .arff, .csv, '
-                          '.jsonlines, .libsvm, .ndj, or .tsv format. '
-                          ' Skipping file {}').format(input_file))
+            logger.error((f"Input file must be in either .arff, .csv, "
+                          f".jsonlines, .libsvm, .ndj, or .tsv format. "
+                          f" Skipping file {input_file}"))
```
mulhod, line 147:
```suggestion
            logger.error(f"Input file must be in either .arff, .csv, "
                         f".jsonlines, .libsvm, .ndj, or .tsv format. "
                         f" Skipping file {input_file}")
```

Issue #624, opened 2020-07-15
Learning curve breaks for very small data sizes


desilinguist:
If you only have a very small amount of training data, learning curves can be very unreliable and may actually raise an exception for very small sizes esp. if you have > 2 labels. 

We should basically have the `learning_curve()` method raise an exception if it's called with less than 500 training examples.

PR #631, opened 2020-10-27
min. number of examples check for the the learning_curve method

srhrshr:
@desilinguist , this PR addresses #624 - I've added a hard minimum number of examples constraint for the `learning_curve` method. Let me know if this is what you had in mind.


skll/learner/__init__.py
```diff
@@ -1618,7 +1618,8 @@ def learning_curve(self,
                        examples,
                        metric,
                        cv_folds=10,
-                       train_sizes=np.linspace(0.1, 1.0, 5)):
+                       train_sizes=np.linspace(0.1, 1.0, 5),
+                       min_training_examples=500):
```
desilinguist, line 1622: Hmm, I am not sure we want to make this an argument? If we know that less than 500 examples lead to unreliable curves, why not simply hardcode that and add an argument like `override_minimum` which is `False` by default but users can choose to override it if they really want to. So, in this case, we would raise a `ValueError` if the number of features doesn't meet the 500 minimum. We would downgrade that error to a warning though if the user sets `override_minimum` to `True`. 

What do you guys think @aoifecahill @mulhod @bndgyawali ?

desilinguist (changes requested):

skll/learner/__init__.py
desilinguist: BTW, whatever we end up doing, we would need to add new tests.

mulhod: Yeah, `override_minimum` sounds good to me.

skll/learner/__init__.py
```diff
@@ -1648,6 +1649,14 @@ def learning_curve(self,
...
             at least one sample from each class.
             Defaults to  ``np.linspace(0.1, 1.0, 5)``.
+        override_minimum : bool, optional
+            Should this be True, learning curve would be generated
+            even with less than ideal number of `examples` (500).
+            However, by default, if the number of `examples` in the FeatureSet is
+            less than 500, an exception is raised, because
+            learning curves can be very unreliable
+            for very small sizes esp. if you have > 2 labels.
+            Defaults to False.
```
desilinguist, line 1659:
```suggestion
        override_minimum : bool, optional
            Learning curves can be unreliable for very small sizes
            esp. for > 2 labels. If this option is set to ``True``, the 
            learning curve would be generated even if the number
            of example is less 500 along with a warning. If ``False``,
            the curve is not generated and an exception is raised instead.
            Defaults to ``False``.
```

skll/learner/__init__.py
```diff
@@ -1658,7 +1667,25 @@ def learning_curve(self,
         num_examples : list of int
             The numbers of training examples used to generate
             the curve
+
+        Raises
+        ------
+        ValueError
+            If the number of training `examples` is less than 500
```
desilinguist, line 1675:
```suggestion
            If the number of examples is less than 500.
```

skll/learner/__init__.py
```diff
@@ -1658,7 +1667,25 @@ def learning_curve(self,
...
         """
+
+        # check that the number of training examples is more than the minimum
+        # needed for generating a reliable learning curve
+        if len(examples) < 500:
+            if not override_minimum:
+                raise ValueError('Number of training examples provided - {} - '
+                                 'is less than the minimum needed - {} - '
+                                 'for the learning curve to be reliable.'.format(len(examples), 500))
```
desilinguist:
```suggestion
                raise ValueError(f"Number of training examples provided({len(examples)}) "
                                 f"is less than the minimum needed(500) "
                                 f"for the learning curve to be reliable.")
```

Since SKLL is Python 3.6+ compatible, it's always better to use the modern f-strings.

skll/learner/__init__.py
```diff
@@ -1658,7 +1667,25 @@ def learning_curve(self,
...
+            else:
+                self.logger.warning('Because the number of training examples provided - {} '
+                                    'is less than the ideal minimum - {} - '
+                                    'learning curve generation is unreliable and might break'.format(len(examples), 500))
+
```
desilinguist, line 1689: Please change this as well.

desilinguist (changes requested):

tests/test_custom_learner.py
```diff
@@ -229,3 +229,23 @@ def test_custom_learner_api_missing_file():
 def test_custom_learner_api_bad_extension():
     other_dir = join(_my_dir, 'other')
     _ = Learner('_CustomLogisticRegressionWrapper', custom_learner_path=join(other_dir, 'custom_learner.txt'))
+
+
+@raises(ValueError)
+def test_custom_learner_learning_curve_min_examples():
+    """
+    test to check that Learner.learning_curve raises an error
+    when the number of training examples provided is less than 500
```
desilinguist, line 238:
```suggestion
    Test to check learning curve raises error with less than 500 examples
```

We don't want these to be too long since they are hard to read when run in verbose mode.

tests/test_custom_learner.py
```diff
@@ -229,3 +229,23 @@ def test_custom_learner_api_missing_file():
...
+    train_fs_less_than_500, _ = make_classification_data(num_examples=499,
+                                                         train_test_ratio=1.0,
+                                                         num_labels=3)
+
+    # creating an example learner
+    learner = Learner('LogisticRegression')
+
+    # this must throw an error because `examples` has less than 500 items
+    _ = learner.learning_curve(examples=train_fs_less_than_500, metric='accuracy',
+                               override_minimum=False)
```
desilinguist, line 251: I think we don't want to specify `override_minimum` here since the goal is to check that the error is raised by default. 

I think we need another tests, where we specify `override_minimum` as `True`, and then check that the warning was output to the log file. I believe we have other tests that already  check for warnings in the log so you should be able to adapt those.

desilinguist (changes requested):

tests/test_custom_learner.py
srhrshr: That makes sense - I've updated this test and have also added a separate test to check that the warning is raised: https://github.com/EducationalTestingService/skll/pull/631/files#diff-d8e9ffeba7d07b6f722f054d5a627f2582ba38b8b90ea090fb38d30cea565936R261-R292

mulhod (approved):
Looks great! I'm not sure why the build has not yet completed, though. Looks like it's stuck in queued status. We might need to manually run it again.

Update: Actually, it does seem to be running now.

tests/test_custom_learner.py
```diff
@@ -229,3 +239,54 @@ def test_custom_learner_api_missing_file():
...
+
+    # checks that the learning_curve warning message is contained in the log file
+    with open(log_file.name) as tf:
+        log_text = tf.read()
+        learning_curve_warning_re = \
+            re.compile(r'Because the number of training examples provided - '
+                       r'\d+ - is less than the ideal minimum - \d+ - '
+                       r'learning curve generation is unreliable'
+                       r' and might break')
+        assert learning_curve_warning_re.search(log_text)
```
desilinguist, line 292: These look great @srhrshr ! One minor thing: is there a reason why these test are in `test_custom_learner.py` rather than `test_output.py` where the other learning curve tests are? These don't have anything to do with custom learners right?

tests/test_custom_learner.py
```diff
@@ -10,17 +10,20 @@
...
 import numpy as np
 from nose.tools import raises
 from numpy.testing import assert_array_equal
+
 from skll.data import NDJWriter
 from skll.experiments import run_configuration
 from skll.learner import Learner
 from skll.utils.constants import KNOWN_DEFAULT_PARAM_GRIDS
-
+from skll.utils.logging import (get_skll_logger)
```
desilinguist, line 26:
```suggestion
from skll.utils.logging import get_skll_logger
```

tests/test_custom_learner.py
```diff
@@ -70,6 +77,9 @@ def tearDown():
                         glob(join(output_dir, 'test_model_custom_learner_*'))):
         os.unlink(output_file)
 
+    for log_file in glob(join(log_dir, '*')):
```
desilinguist, line 80: I think rather than deleting all log files which could affect other tests, we should only delete the specific log file we are actually generating.

desilinguist (changes requested):
We are almost there @srhrshr ! Thanks for being so receptive to our suggestions 🙇🏽.

tests/test_custom_learner.py
srhrshr: That's a good point - I didn't realize there were learning curve tests as I was looking for tests of the `Learner*` class all this while 🤦   Will move the tests to `test_output.py`

srhrshr: Done @desilinguist !

tests/test_custom_learner.py
srhrshr: Done! I've also done a little bit of cleaning up to the tearDown function and instead of creating a new log dir, I've written the log file to an `output` dir, since the log files are being written to the `output` dir in a few other places.

desilinguist: Thanks for doing this. One minor nit I want to pick is that I am not a fan of single letter variable names and nested list comprehensions when they aren't necessary and sacrifice readability. I think in this case, nested for loops would work just fine.

srhrshr: I totally get that - this practice of mine has always been a little polarizing :P 
Made this change.

desilinguist: Thank you!

desilinguist:
Seems like a missing import?

desilinguist:
Hmm, I am seeing a [partial failure](https://dev.azure.com/EducationalTestingService/SKLL/_build/results?buildId=967&view=logs&j=54de5b39-23c3-5288-53e4-8fc282b5ee50&t=c73320f3-a9bb-5060-2ee4-6ee4065e03c9) on the windows builds:

tests/test_output.py
```diff
@@ -89,8 +88,7 @@ def tearDown():
                 os.unlink(join(config_dir, cf))
 
         for output_file in glob(join(output_dir, 'test_{}_*'.format(suffix))) \
-            + glob(join(output_dir, 'test_{}.log'.format(suffix))) \
-                + glob(join(output_dir, 'test_majority_class_custom_learner_*')):
+                           + glob(join(output_dir, 'test_{}.log'.format(suffix))):
```
mulhod, line 91: It looks like this isn't a `glob` pattern? I think it could be dealt with elsewhere. I'm not sure if this has anything to do with the build failure, but it's probably better to remove it above (or below) if it exists.

tests/test_output.py
```diff
@@ -894,6 +895,57 @@ def test_learning_curve_ylimits():
...
+
+def test_learning_curve_min_examples_check_override():
+    """
+    Test to check learning curve displays warning with less than 500 examples
+    """
+
+    # creates a logger which writes to a temporary log file
+    log_file_path = join(_my_dir, "output",
+                         "test_learning_curve_min_examples_check_override.log")
+    logger = get_skll_logger("test_learning_curve_min_examples_check_override",
```
mulhod, line 923: Once this logger is no longer need, it needs to be closed (and the file handle needs to be closed). Use this function to do that:

```python
from skll.utils.logging import close_and_remove_logger_handlers

close_and_remove_logger_handlers(logger)
```

This would be done at the end of the test (after `learner.learning_curve` is called at least).

mulhod (changes requested):
The file handle thing needs to be fixed. It's causing the build to fail.

tests/test_output.py
srhrshr: I don't think this was the issue. But line 90 (`'test_{}_*'.format(suffix)))`) was also matching the log filename of my test, despite me trying to be super specific about the filename I used. I've changed it now. 

Btw, is the `tearDown` method called after each test? Because, that would explain the race condition.

desilinguist: Yes, I believe that's right.

mulhod (approved):
LGTM!

desilinguist (approved):
Finally! Looks great! Excellent work @srhrshr !

PR #632, opened 2020-10-27
updates scikit to 0.23.2

### LINKED ISSUE #630: Update scikit-learn to v0.23.2
According to the [changelog](https://scikit-learn.org/stable/whats_new/v0.23.html#version-0-23-2), there aren't any major changes so should be an easy update.

srhrshr:
@desilinguist , this closes #630. Will update this PR with local nose test results to double check that there aren't any regressions.

### DIFF
--- a/conda-recipe/skll/meta.yaml
+++ b/conda-recipe/skll/meta.yaml
@@ -39,7 +39,7 @@ requirements:
     - pandas
     - python >=3.6
     - ruamel.yaml
-    - scikit-learn ==0.23.1
+    - scikit-learn ==0.23.2
     - scipy
     - seaborn
     - setuptools
@@ -52,7 +52,7 @@ requirements:
     - pandas
     - python >=3.6
     - ruamel.yaml
-    - scikit-learn ==0.23.1
+    - scikit-learn ==0.23.2
     - seaborn
     - scipy
     - tabulate
--- a/conda-recipe/skll/meta.yaml
+++ b/conda-recipe/skll/meta.yaml
@@ -39,7 +39,7 @@ requirements:
     - pandas
     - python >=3.6
     - ruamel.yaml
-    - scikit-learn ==0.23.2
+    - scikit-learn >=0.23.1,<=0.23.2
     - scipy
     - seaborn
     - setuptools
@@ -52,7 +52,7 @@ requirements:
     - pandas
     - python >=3.6
     - ruamel.yaml
-    - scikit-learn ==0.23.2
+    - scikit-learn >=0.23.1,<=0.23.2
     - seaborn
     - scipy
     - tabulate
--- a/conda_requirements.txt
+++ b/conda_requirements.txt
@@ -3,7 +3,7 @@ joblib>=0.8
 numpy
 pandas
 ruamel.yaml
-scikit-learn==0.23.1
+scikit-learn==0.23.2
 scipy
 seaborn
 tabulate
--- a/examples/make_titanic_example_data.py
+++ b/examples/make_titanic_example_data.py
@@ -10,7 +10,13 @@ titanic.cfg expects.
 
 import logging
 import os
+import sys
 from itertools import chain
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+sys.path.insert(0, os.path.abspath('../../'))
 
 from skll.data import Reader, Writer
 
--- a/examples/make_titanic_example_data.py
+++ b/examples/make_titanic_example_data.py
@@ -10,13 +10,7 @@ titanic.cfg expects.
 
 import logging
 import os
-import sys
 from itertools import chain
-
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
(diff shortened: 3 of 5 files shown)


srhrshr:
Ran into 1 `nosetests` error due to a module root issue, and the sys.path insertion commit fixes it.  All good now. 
```
Traceback (most recent call last):
  File "--------------/skll/examples/make_titanic_example_data.py", line 15, in <module>
    from skll.data import Reader, Writer
ModuleNotFoundError: No module named 'skll'
Retrieving boston data from servers...done
Writing training and testing files...done
Retrieving iris data from servers...done
Writing training and testing files...done
E
======================================================================
ERROR: Run all of the configuration files for the titanic example
----------------------------------------------------------------------
Traceback (most recent call last):
  File "........../.pyenv/versions/3.6.9/envs/skll-0.23.2/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "...............skll/tests/test_examples.py", line 133, in test_titanic_configs
    run_configuration_and_check_outputs(config_path)
  File "...............skll/tests/test_examples.py", line 48, in run_configuration_and_check_outputs
    results_json_path = run_configuration(config_path, local=True, quiet=True)[0]
  File "...............skll/skll/experiments/__init__.py", line 570, in run_configuration
    log_level=log_level)
  File "...............skll/skll/config/__init__.py", line 628, in parse_config_file
    train_path = locate_file(train_path, config_dir)
  File "...............skll/skll/config/utils.py", line 106, in locate_file
    raise IOError(errno.ENOENT, "File does not exist", path_to_check)
FileNotFoundError: [Errno 2] File does not exist: '...............skll/tests/other/titanic/train+dev' ```

File: examples/make_titanic_example_data.py [hunk 1]
```diff
@@ -10,8 +10,14 @@
 
 import logging
 import os
+import sys
 from itertools import chain
 
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+sys.path.insert(0, os.path.abspath('../../'))
```
desilinguist, line 19: This should not be necessary @srhrshr. Did you run `pip install -e .` before running the tests locally?

desilinguist (changes requested):
Thanks for taking this on @srhrshr! One other thing I check is whether there are any underlying warnings from sklearn in case they are about to deprecate something or have moved things around? Since you are running the tests locally, can you also check for those warnings please? 

Also, I don't think the sys.path change should be necessary.

conda-recipe/skll/meta.yaml
```diff
@@ -39,7 +39,7 @@ requirements:
     - pandas
     - python >=3.6
     - ruamel.yaml
-    - scikit-learn ==0.23.1
+    - scikit-learn ==0.23.2
```
mulhod, line 42: Since it appears `0.23.1` should still work, can we relax this to the following?

```suggestion
    - scikit-learn >=0.23.1,<=0.23.2
```

requirements.txt
```diff
@@ -3,7 +3,7 @@ joblib>=0.8
 numpy
 pandas
 ruamel.yaml
-scikit-learn==0.23.1
+scikit-learn==0.23.2
```
mulhod, line 6: Same as above, please relax to:

```suggestion
scikit-learn>=0.23.1,<=0.23.2
```

conda_requirements.txt
```diff
@@ -3,7 +3,7 @@ joblib>=0.8
 numpy
 pandas
 ruamel.yaml
-scikit-learn==0.23.1
```
mulhod, line 6: FYI: While I'm suggesting relaxing the requirement in a couple other places, _here_ I don't think anything needs to be done since this is just for development dependencies.

mulhod (changes requested):

mulhod:
@srhrshr Thanks for the PR, by the way! It's much appreciated. I just had a couple comments about relaxing the requirement so that it includes `0.23.2` but doesn't exclude `0.23.1`.

examples/make_titanic_example_data.py
srhrshr: @desilinguist - yep that fixes it! I was doing `pip install -r requirements.txt`. Thanks! Reverting this change.

conda-recipe/skll/meta.yaml
srhrshr: That makes sense @mulhod. Updated.

tests/test_metrics.py
```diff
@@ -122,7 +122,7 @@ def check_f05_metrics(metric_name, average_method):
     y_true = [1, 1, 1, 0, 0, 0]
     y_pred = [0, 1, 1, 1, 0, 0]
     skll_value = use_score_func(metric_name, y_true, y_pred)
-    sklearn_value = fbeta_score(y_true, y_pred, 0.5, average=average_method)
```
srhrshr, line 125: @desilinguist , this is a change that addresses a warning error I got during local tests:
```FutureWarning: Pass beta=0.5 as keyword args. From version 0.25 passing these as positional arguments will result in an error```
Otherwise, there was nothing that sklearn shouted which was deprecation related.

mulhod (approved):

desilinguist (approved):
This looks great @srhrshr !  Great work!

PR #634, opened 2020-11-09
Use f-strings instead of format or old-style %

mulhod:


mulhod:
According to the code coverage diff, coverage increases very slightly by 0.01%. So, why `codecov` is saying differently I'm not really sure.

desilinguist:
@mulhod can you push another commit?

mulhod:
Hmm, `codecov` check still failed.

mulhod:
Or perhaps I'm just reading it wrong? The `codecov/patch` check failed. But not `codecov/project`.

desilinguist:
Yeah, that's the patch diff that failed – which checks whether any new lines you added are also covered by tests. I am guessing these are because you broke some long strings into smaller ones?

mulhod:
I did break some lines up, but it appears the total didn't check by much. Maybe 2 lines more?

Anyway, it's a little unclear to me how this can be addressed or if it should.

desilinguist:
I don't think we need to worry about it.

desilinguist (approved):
LGTM 👍🏽 !

bndgyawali (approved):
I have one small suggestion, else it LGTM.

skll/data/readers.py
```diff
@@ -280,7 +280,7 @@ def feat_dict_generator():
                 for ex_num, (_, _, feat_dict) in enumerate(self._sub_read(f)):
                     yield feat_dict
                     if ex_num % 100 == 0:
-                        self._print_progress('{100 * (ex_num / total):.8}%')
+                        self._print_progress('{100 * ex_num / total:.8}%')
```
desilinguist, line 283: Does this actually work?

mulhod: Yes, it does appear to.

PR #643, opened 2020-11-13
Add closing of logger file handlers to tearDown method in test_input.py

mulhod:


mulhod:
Let's see if this works.

desilinguist (approved):
LGTM!  👍

desilinguist:
I think we can merge this without a second review given the very limited scope of the bugfix and the fact that the broken build now passes.

PR #644, opened 2020-11-16
Fix up some cases where open was not done with a context manager, etc.

### LINKED ISSUE #641: Use with context manager when opening files
In some cases, `open`ing a file is not done with a context manager.

Furthermore, `"r"` is assumed and can be removed wherever it's used pretty much.

mulhod:
Closes #641.


desilinguist (approved):
LGTM 👍!

bndgyawali (approved):
LGTM

PR #645, opened 2020-11-18
Fix up some string concatenation, line length issues, etc.

### LINKED ISSUE #636: Standardize string concatenation
Standardize string concatenation over multiple lines.

There are cases where we're concatenating with +, with inline /, and where we needlessly parenthesize the whole string (for example, in an argument to a function).

mulhod:
Closes #636 

Mainly, this PR has to do with cleaning up some multi-line string concatenations (or removing them altogether when possible). There are a couple other randomish improvements I made as well (like reducing the length of some very long lines or lines where it was easy to do, removing unused parameters from a function used by a test, etc.).


mulhod:
Why the coverage check is failing I have no idea. When I click on the "Details" link, I don't see any change in coverage... Perhaps I'm not reading it correctly?

desilinguist (approved):
Looks great! Thanks for taking this on @mulhod!

mulhod:
@bndgyawali If you get a chance, could you take a look at this? There are quite a few changes, but they are all cosmetic.

bndgyawali (approved):

Issue #646, opened 2020-11-25
Consider using pre-commit hooks for code formatting/problems etc


mulhod:
https://pre-commit.com/

This possibly intersects with #639 since `black` could be hooked in.

desilinguist:
Addressed by #650.

PR #650, opened 2020-12-21
Introduce pre-commit hooks

mulhod:
Hooks are installed as a development dependency, so this shouldn't affect anything in terms of packaging.

I have left out `black` for now because I don't want this getting too big and controversial. I will do that next separately so that we can all see what exactly will happen.


mulhod:
Addresses #646

desilinguist (commented):
Should we exclude the CSV/TSV/JSON files from the test data from the auto-formatting fixes?

desilinguist (changes requested):
Everything else looks good to me! This will be quite useful! Just that one minor suggestion.

mulhod:
> Should we exclude the CSV/TSV/JSON files from the test data from the auto-formatting fixes?

I have gone ahead and excluded these files from the EOF file hook.

desilinguist (approved):
LGTM!

skll/learner/__init__.py
```diff
@@ -11,78 +11,83 @@
...
-                              RandomForestRegressor)
-from sklearn.feature_extraction import FeatureHasher
+from sklearn.dummy import DummyClassifier, DummyRegressor  # noqa: F401
+from sklearn.ensemble import (
+    AdaBoostClassifier,
+    AdaBoostRegressor,
+    GradientBoostingClassifier,
+    GradientBoostingRegressor,
+    RandomForestClassifier,
+    RandomForestRegressor,
```
bndgyawali, line 29: do we need a comma at the end?

skll/learner/__init__.py
```diff
@@ -11,78 +11,83 @@
...
+    Lars,
+    Lasso,
+    LinearRegression,
+    LogisticRegression,
+    RANSACRegressor,
+    Ridge,
+    RidgeClassifier,
+    SGDClassifier,
+    SGDRegressor,
+    TheilSenRegressor,
```
bndgyawali, line 47: here too.

skll/learner/__init__.py
```diff
@@ -11,78 +11,83 @@
...
+    compute_num_folds_from_example_counts,
+    get_acceptable_classification_metrics,
+    get_acceptable_regression_metrics,
+    get_predictions,
+    load_custom_learner,
+    rescaled,
+    setup_cv_fold_iterator,
+    setup_cv_split_iterator,
+    train_and_score,
+    write_predictions,
```
bndgyawali, line 89: and here

skll/learner/utils.py
```diff
@@ -22,21 +21,27 @@
...
-                             confusion_matrix,
-                             precision_recall_fscore_support)
-from sklearn.model_selection import (KFold,
-                                     LeaveOneGroupOut,
-                                     ShuffleSplit,
-                                     StratifiedKFold)
+from sklearn.metrics import (
+    accuracy_score,
+    confusion_matrix,
+    precision_recall_fscore_support,
```
bndgyawali, line 27: seems it is having comma everywhere in this type of import statements.

bndgyawali (approved):
I have a comment, but seems like that is how black does. 
Else everything looks good to me.

aoifecahill (approved):

Issue #656, opened 2021-02-03
`Learner._check_input_formatting()` does not work for dense featuresets


desilinguist:
This method is called by `Learner._train_setup()` and it checks that regression labels are not strings and that feature values (for both classification and regression) are not strings. However, this method does not work as expected if the featureset is read in as dense rather than sparse. Here's a minimal test case:

```python
>>> from skll.data import NDJReader
>>> fs1 = NDJReader.for_path("examples/boston/train/example_boston_features.jsonlines", sparse=False).read()
>>> l1 = Learner('LinearRegression')
>>> fs2 = NDJReader.for_path("examples/iris/train/example_iris_features.jsonlines", sparse=False).read()
>>> l2 = Learner('LogisticRegression')
>>> l1.train(fs1, grid_search=False)
...
~/work/skll/skll/learner/__init__.py in _check_input_formatting(self, examples)
    664         # make sure that feature values are not strings
    665         # we need to check this for both sparse and dense arrays
--> 666         for val in examples.features.data:
    667             if isinstance(val, str):
    668                 raise TypeError("You have feature values that are strings.  "

NotImplementedError: multi-dimensional sub-views are not implemented

>>> l2.train(fs2, grid_search=False)
....
~/work/skll/skll/learner/__init__.py in _check_input_formatting(self, examples)
    664         # make sure that feature values are not strings
    665         # we need to check this for both sparse and dense arrays
--> 666         for val in examples.features.data:
    667             if isinstance(val, str):
    668                 raise TypeError("You have feature values that are strings.  "

NotImplementedError: multi-dimensional sub-views are not implemented
```

The solution is to explicitly reshape the dense feature array into a 1-dimensional array before iterating over `.data` attribute.

PR #659, opened 2021-02-03
Update scikit-learn to 0.24.1

### LINKED ISSUE #653: Upgrade scikit-learn to 0.24.1
See full [changelog](https://scikit-learn.org/stable/whats_new/v0.24.html#version-0-24-0). Since this has many API changes, this might be a bit of work.

desilinguist:
This PR closes #653. 

It pretty much works out of the box except for two changes:

- Setting the new keyword argument `error_score` to "raises" in the GridSearchCV() call made in Learner.train() since we want to raise an exception if there was any problem with fitting the estimator. This change is necessary because the new scikit-learn default is to simply return a nan as the fit score in case of a problem which does not work for us. 

- `LinearRegression` models in scikit-learn now support a new keyword argument `positive` which can be set to `True` to use Non-negative Least Squares (NNLS) regression. This is probably something we want to enable in SKLL since it could be useful in [RSMTool](https://github.com/EducationalTestingService/rsmtool). This required the fix for #656 which has already been merged. 

- Add a test for this new non-negative regression. 

- Other minor changes:
  + Since Python 3.6 is so long in the tooth, I have changed the Linux builds (on Travis) to use Python 3.7 and the Windows builds (on Azure) to use Python 3.8. I am using Python 3.9 locally.

  + Update both `requirements.txt` and `conda_requirements.txt` to use the new version of scikit-learn.


aoifecahill (approved):

mulhod (approved):
Awesome!

PR #660, opened 2021-02-04
Fix bugs in `filter_features` command line utility.

### LINKED ISSUE #598: Two bugs in `filter_features`
There are two issues with `filter_features`:

1. Any labels specified with `-L` or `--label` are assumed to be strings. This will not always work because when reading in feature file, we will convert labels to floats or ints whenever possible.  So, if the labels are, for example, 0 and 1, then `filter_features -L 0 1` will never work.

2. The argument parsing for `filter_features` is strange because several options have `nargs='*'` which makes it so that the input file and output file have to be specified _first_ before any of these options. 

The fixes for both these issues are simple:

1. We should be calling `safe_float()` on the labels parsed in `filter_features` so that these labels are converted in the exact same way as the labels when the feature file is read in. 

2.  We need to switch to using `nargs='+'` and use the `required` keyword argument.

Another thing to note that since 1 was never caught with current tests, we have a gap in our `filter_features` tests.

desilinguist:
This PR closes #598.

- Run any labels specified via `-L` through `safe_float()` to make sure that they are converted in the same way that the FeatureSet reader classes do for any feature files (see issue for more details).
- Change the command line arguments for `filter_features` to make them more intuitive to newcomers. This breaks backwards compatibility but it's worth it. Specifically, input and output files must be specified using `-i` and `-o` respectively, and inversion is now effected by using `--inverse`. 
- Update existing `filter_feature` tests to use the new command line arguments.
- Add new tests for the label conversion. 
- Update the `filter_features` documentation with the new arguments as well as by adding a new warning.

### DIFF
--- a/doc/utilities.rst
+++ b/doc/utilities.rst
@@ -48,14 +48,24 @@ filter_features
 Filter feature file to remove (or keep) any instances with the specified IDs or
 labels.  Can also be used to remove/keep feature columns.
 
-Positional Arguments
-^^^^^^^^^^^^^^^^^^^^
-.. option:: infile
+.. warning:: Starting with v2.5 of SKLL, the arguments for ``filter_features``
+             have changed and are no longer backwards compatible with older
+             versions of SKLL. Specifically:
+
+             #. The input and output files must now be specified with ``-i``
+                and ``-o`` respectively
+
+             #. ``--inverse`` must now be used to invert the filtering command
+                since ``-i`` is used to specify the input file.
+
+Required Arguments
+^^^^^^^^^^^^^^^^^^
+.. option:: -i, --input
 
     Input feature file (ends in ``.arff``, ``.csv``, ``.jsonlines``,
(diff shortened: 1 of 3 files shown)


aoifecahill (approved):

doc/utilities.rst
```diff
@@ -48,14 +48,24 @@ filter_features
...
 
-Positional Arguments
-^^^^^^^^^^^^^^^^^^^^
-.. option:: infile
+.. warning:: Starting with v2.5 of SKLL, the arguments for ``filter_features``
+             have changed and are no longer backwards compatible with older
+             versions of SKLL. Specifically:
+
+             #. The input and output files must now be specified with ``-i``
+                and ``-o`` respectively
```
mulhod, line 56: Nitpick:

```suggestion
                and ``-o`` respectively.
```

mulhod (approved):
Looks good to me!

One very minor editing comment. Address if you have the inclination.

PR #662, opened 2021-02-05
Remove top-level imports

### LINKED ISSUE #661: Remove top-level imports
As I work on #488, I find that one unfortunate by-product of adding a new `VotingLearner` object that depends on the `Learner` object  is that the following no longer works: `from skll import Learner`  due to a circular import issue. From v2.5, we will need to do `from skll.learner import Learner` instead . 

To make things consistent, I plan to remove all top-level imports:
-  `from skll import FeatureSet` will become `from skll.data import FeatureSet`  
- `from skll import run_configuration` will become `from skll.experiments import run_configuration`

Since we are already planning to break backwards compatibility with v2.5, the change can be clearly mentioned in the release notes. Also, IMO, it provides the user with a better sense of the structure of the SKLL codebase.

desilinguist:
This PR closes #661. 

In addition, it makes another minor change of passing in the `beta` argument as a keyword argument since scikit-learn raises a warning otherwise.

### DIFF
--- a/skll/__init__.py
+++ b/skll/__init__.py
@@ -11,12 +11,7 @@ common scikit-learn experiments with pre-generated features.
 
 from sklearn.metrics import SCORERS, f1_score, fbeta_score, make_scorer
 
-from .data import FeatureSet
-from .experiments import run_configuration
-from .learner import Learner
 from .metrics import correlation, f1_score_least_frequent, kappa
-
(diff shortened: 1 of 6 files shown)


desilinguist:
The codecov change is expected since we are removing import lines.

aoifecahill (approved):

mulhod (approved):

PR #663, opened 2021-02-05
py36-compatible V2.0 branch

slava92:
This is mostly for record-keeping. I built skll=2.0.1 conda package for py38 and placed in ETS internal channel


desilinguist:
Yep, I am going to close this.

PR #667, opened 2021-02-22
Use 5 folds for grid search

### LINKED ISSUE #487: Change default grid search folds
The scikit-learn folks are planning to change the default number of CV splits for all kinds of cross-validation splitters from 3 to 5 since 3 has too much variance. 

Should we also do this in for our default `grid_search_fold` and `SKLL_MAX_CONCURRENT_PROCESSES` values?  This would mean our SGE workflows would change to require 5 slots instead of 3 whenever we run on the grid. However, as we have also seen, variance among the folds is bad.

desilinguist:
This PR closes #487. 

See issue for motivation to change the default number of folds for grid search cross-validation. One by-product of the change is that experiments will certainly take longer to run but they are likely to be more reliable. 

Changes: 
- Change default grid search folds from 3 to 5. 
- Update `grid_search_folds` documentation
- Change default value of `MAX_SKLL_CONCURRENT_PROCESSES` to 5.
- Change tests to use 3 grid search folds otherwise tests take too long. Note that the the config parsing tests in `test_input.py` were correctly modified to check that the default value for grid search folds is 5.


desilinguist:
@mulhod it'd be really great if you could test this with gridmap as well. Thanks!

aoifecahill (approved):
Looks good, thanks!

mulhod (approved):

PR #679, opened 2021-03-10
Replace MANIFEST.in with `package_data` in setup.py

### LINKED ISSUE #674: Replace MANIFEST.in with `package_data` in setup.py
We shouldn't need to use a separate file to include additional files anymore since latest versions of `setuptools` include [data files support](https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html).

RobertImbrie:
This PR closes #674.
- Add package_data, exclude_data, and include_package_data methods to setup.py
- Delete MANIFEST.in

**Note:** Tested by running `python setup.py sdist build` once for each setup and running a diff on the generated folders. They were identical!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include LICENSE.txt
+include requirements.txt
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,0 @@
-include LICENSE.txt
-include *.rst
-include requirements.txt
-recursive-exclude doc *
-recursive-exclude tests *
-recursive-exclude examples *
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,10 @@ setup(name='skll',
       author_email='nmadnani@ets.org',
       license='BSD 3 clause',
       packages=find_packages(exclude=['tests', 'examples']),
+      package_data={'':
+                    ['*.rst', 'LICENSE.txt', 'requirements.txt']},
+      exclude_package_data={'doc': ['*'], 'tests': ['*'], 'examples': ['*']},
+      include_package_data=True,
       entry_points={'console_scripts':
                     ['filter_features = skll.utils.commandline.filter_features:main',
                      'generate_predictions = skll.utils.commandline.generate_predictions:main',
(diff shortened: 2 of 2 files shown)


RobertImbrie:
Thank you, Nitin! I'm hoping to keep them going. :)

Also, glad to hear it went through easily this time!

desilinguist:
So, I have been playing around with this and I don't think these changes are giving us the desired result. Here are the things I tried:

1. If I remove `MANIFEST.in` and run `python setup.py sdist build`, `tests` and `examples` are correctly excluded but `LICENSE.TXT` and `requirements.txt` are *not* included. 

2. If I modify the `find_packages()` call in `setup.py` to not mention `exclude=['tests', 'examples']`  and run `python setup.py sdist build`,  `tests` and `examples` are _not_ excluded this time around and LICENSE.txt and requirements.txt are still _not_ included. 

3. If I modify `setup.py` to now add back `exclude=['tests', 'examples']` and then also add the following two lines to it, `tests` and `example` are now back to being correctly excluded but `LICENSE.txt` and `requirements.txt` are _still_ not included:
    ```python
    package_data={'': ['*.rst', 'LICENSE.txt', 'requirements.txt']},
    include_package_data=True,
    ```

So, I referred to the [relevant documentation](https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html?highlight=package_data) for `setuptools` and found this parenthetical nugget:

> Note: although the package_data argument was previously only available in setuptools, it was also added to the Python distutils package as of Python 2.4; there is some documentation for the feature available on the python.org website. If using the setuptools-specific include_package_data argument, files specified by package_data will not be automatically added to the manifest unless they are listed in the MANIFEST.in file.

So, what I tested next was to keep `MANIFEST.in` but to edit it to only keep the following lines: 
```
include LICENSE.txt
include requirements.txt
```

And this yielded exactly the same results as what we have right now (`README.rst` is already included so we don't need `include *.rst` either)

@RobertImbrie can you please edit this PR to make the following changes:
- Add back `MANIFEST.in` with the same two files.
- Revert any changes to `setup.py`.

desilinguist (changes requested):
See my comment for suggested changes.

RobertImbrie:
Hi Nitin! I made the requested changes to the pull request.

Also, my sincere apologies for the issues! I had deleted the `build` directory but _not_ `skll.egg-info`, which should account for why my output was different. Next time, I will read the documentation more closely.

desilinguist:
Looks great! Thanks @RobertImbrie!

desilinguist (approved):

Issue #647, opened 2020-11-25
Rewrite docstrings to match scikit-learn style and show type hints correctly.


desilinguist:
We should try and match the way that scikit-learn writes their docstrings. [Here](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/_huber.py#L125) is an example. 

For example, there is a blank line between the parameters and the default values are indicated right after the parameter type. It's also nice how the types that are numpy arrays also include the shapes.

RobertImbrie:
Hey @desilinguist! Finally have some time to get started on this again.

I noticed there's inconsistencies between how we write boolean docstrings and how scikit writes them. Even worse, scikit itself is inconsistent with how it writes boolean docstrings!

I want to make our documentation for booleans consistent. That leaves with us three options:

**skll's current method:** Should we do X or not?
**scikit method 1:** If ``True``, do X.
**scikit method 2:** Whether or not to do X.

I'm leaning towards scikit method 1, though I also don't want to make any unilateral changes. What are your thoughts?

desilinguist:
Welcome back @RobertImbrie! We really appreciate your contributions. 

That's a very interesting finding regarding Boolean docstrings! I think I also agree with your thinking and like scikit option 1 - the most important thing is consistency. 

What do you think @mulhod @Frost45  @aoifecahill?

mulhod:
Yeah, sounds good. I agree about merely being consistent.

PR #681, opened 2021-03-18
Rewrite docstrings for skll/config/__init__.py

RobertImbrie:
Addresses issue #647.
- Add line of spacing between docstring parameters.
- Indicate default values after parameter type.
- Add missing period to a single parameter's description.

**Note:** No array parameters found in this file.

### DIFF
--- a/skll/config/__init__.py
+++ b/skll/config/__init__.py
@@ -260,8 +260,8 @@ def parse_config_file(config_path, log_level=logging.INFO):  # noqa: C901
 
     log_level : int, default=logging.INFO
         Determines which messages should be logged. You can either pass
-        an integer or one of the corresponding `logging` parameters, such
-        as `logging.INFO` or `logging.WARNING`.
+        an integer or one of the corresponding ``logging`` parameters, such
+        as ``logging.INFO`` or ``logging.WARNING``.
 
     Returns
     -------
--- a/skll/config/__init__.py
+++ b/skll/config/__init__.py
@@ -258,8 +258,10 @@ def parse_config_file(config_path, log_level=logging.INFO):  # noqa: C901
     config_path : str
         The path to the configuration file.
 
-    log_level : logging level, optional, default=logging.INFO
-        The logging level to use.
+    log_level : int, default=logging.INFO
+        Determines which messages should be logged. You can either pass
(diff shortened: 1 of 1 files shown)


skll/config/__init__.py
```diff
@@ -254,122 +257,168 @@ def parse_config_file(config_path, log_level=logging.INFO):  # noqa: C901
     ----------
     config_path : str
         The path to the configuration file.
-    log_level : logging level, optional
+
+    log_level : logging level, optional, default=logging.INFO
```
desilinguist, line 261:
```suggestion
    log_level : logging level, default=logging.INFO
```

We should change to an actual type, not just "logging level" and since we are specifying `default` now, we don't need the word "optional" any more. I think that should be implied.

skll/config/__init__.py
```diff
@@ -873,9 +922,9 @@ def _setup_config_parser(config_path, validate=True):
     ----------
     config_path : str
         The path to the configuration file.
-    validate : bool, optional
+
+    validate : bool, optional, default=True
```
desilinguist, line 926:
```suggestion
    validate : bool, default=True
```

desilinguist (changes requested):
Looks great @RobertImbrie! Just some minor suggestions.

skll/config/__init__.py
RobertImbrie: Makes sense! Technically, `log_level` is an integer. However, it's more common to pass `log_level` one of the `logging` parameters, like `logging.INFO` or `logging.WARNING`. I think that might be worth including in the description. Here's an example:

```
    log_level : int, default=logging.INFO
        Determines which messages should be logged. You can either pass
        an ``int`` or one of the corresponding ``logging`` parameters, such
        as ``logging.INFO`` or ``logging.WARNING``. 
```

What are your thoughts?

desilinguist: Great idea! Let's do it!

RobertImbrie:
@desilinguist Suggested changes have been made! Had to make a third commit to fix a typo.

desilinguist (approved):
LGTM!  👍

mulhod (approved):

PR #682, opened 2021-03-22
Rewrite docstrings for skll/config/utils.py

RobertImbrie:
- Add space between parameters.
- Remove 'optional' from parameters with default values.
- Move default values next to parameter names.
- Add missing periods to parameter descriptions.
- Add missing double backtick to variable and object names.
- Fix incorrect variable/object reference in _munge_featureset_name docstrings.
    - The docstrings reference `feature_set`, which does not exist. The function is passed a variable called `featureset`. The `FeatureSet` object is spelled with camel case.
    - As it turns out, `feature_set` was used interchangeably for both the `featureset` variable *and* the `FeatureSet` object.
    - `feature_set` reverted to either `featureset` or `FeatureSet`, as appropriate.
- Shortened _munge_featureset_name function description by just a couple words.
    
**Note:** No array parameters this time either!

### DIFF
--- a/skll/config/utils.py
+++ b/skll/config/utils.py
@@ -111,7 +111,7 @@ def locate_file(file_path, config_dir):
 
 def _munge_featureset_name(featureset):
     """
-    Joins features in ``featureset`` by '+' if ``featureset`` is not a string.
+    Joins features in ``featureset`` with a '+' if ``featureset`` is not a string.
     Otherwise, returns ``featureset``.
 
     Parameters
--- a/skll/config/utils.py
+++ b/skll/config/utils.py
@@ -45,9 +45,9 @@ def load_cv_folds(folds_file, ids_to_floats=False):
     ----------
     folds_file : str
         The path to a folds file to read.
-    ids_to_floats : bool, optional
+
+    ids_to_floats : bool, default=False
         Whether to convert IDs to floats.
-        Defaults to ``False``.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


skll/config/utils.py
```diff
@@ -110,18 +111,18 @@ def locate_file(file_path, config_dir):
 
 def _munge_featureset_name(featureset):
     """
-    Joins features in featureset by '+' if featureset is not a string, and just
-    returns featureset otherwise.
+    Joins features in ``featureset`` by '+' if ``featureset`` is not a string.
```
desilinguist, line 114:
```suggestion
    Joins features in ``featureset`` with a '+' if ``featureset`` is not a string.
```

skll/config/utils.py
```diff
@@ -110,18 +111,18 @@ def locate_file(file_path, config_dir):
...
+    Otherwise, returns ``featureset``.
 
     Parameters
     ----------
     featureset : SKLL.FeatureSet
```
desilinguist, line 119:
```suggestion
    featureset : skll.data.FeatureSet
```

desilinguist (changes requested):
Thanks @RobertImbrie ! A couple of suggestions.

RobertImbrie:
@desilinguist Suggestions added!

desilinguist (approved):
LGTM!

mulhod (approved):

PR #686, opened 2021-04-01
Rewrite docstrings for skll/data/readers.py

RobertImbrie:
- Add space between parameters.
- Change ` to `` when referring to code.
- Remove 'optional' from parameters with default values.
- Move default values next to parameter names.
- Specify shape of arrays and array-likes.
- Update `skll.FeatureSet` and `skll.Reader` to `skll.data.FeatureSet` and `skll.data.Reader`
- Line up variables and text in `replace_blanks_with` parameter descriptions to improve legibility.

**Notes:**
Before the pull request is approved, I'd like to note that both instances of the `replace_blanks_with` parameter include a 'value' option with unspecified type, shown here:
```
replace_blanks_with : value, ``dict``, or ``None``, default=None
```

Later, in the parameter description, it says that this value must be numeric:
```
value   = A (numeric) value with which to replace blank values.
```

If that's the case, shouldn't we refer to it as `Number` instead of 'value'?

### DIFF
--- a/skll/data/readers.py
+++ b/skll/data/readers.py
@@ -67,48 +67,48 @@ class Reader(object):
     ----------
     path_or_list : str or list of dict
         Path or a list of example dictionaries.
-    quiet : bool, optional
+
+    quiet : bool, default=True
         Do not print "Loading..." status message to stderr.
-        Defaults to ``True``.
-    ids_to_floats : bool, optional
+
+    ids_to_floats : bool, default=False
         Convert IDs to float to save memory. Will raise error
         if we encounter an a non-numeric ID.
-        Defaults to ``False``.
-    label_col : str, optional
+
+    label_col : str, default='y'
         Name of the column which contains the class labels
         for ARFF/CSV/TSV files. If no column with that name
         exists, or ``None`` is specified, the data is
         considered to be unlabelled.
-        Defaults to ``'y'``.
-    id_col : str, optional
+
+    id_col : str, default='id'
         Name of the column which contains the instance IDs.
         If no column with that name exists, or ``None`` is
         specified, example IDs will be automatically generated.
-        Defaults to ``'id'``.
-    class_map : dict, optional
+
+    class_map : dict, default=None
         Mapping from original class labels to new ones. This is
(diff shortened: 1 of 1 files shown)


skll/data/readers.py
```diff
@@ -695,25 +710,26 @@ class CSVReader(Reader):
...
-        The replacement occurs after the data set is read into a `pd.DataFrame`.
-        Defaults to ``None``.
-    drop_blanks : bool, optional
+    drop_blanks : bool, default=False
         If ``True``, remove lines/rows that have any blank
         values. These lines/rows are removed after the
-        the data set is read into a `pd.DataFrame`.
-        Defaults to ``False``.
-    pandas_kwargs : dict or None, optional
+        the data set is read into a ``pd.DataFram`e`.
```
desilinguist, line 727:
```suggestion
        the data set is read into a ``pd.DataFrame``.
```

desilinguist (changes requested):
Thanks @RobertImbrie ! Found a typo and also seeing some other issues that it's hard to show suggestions for. Will push a commit.

desilinguist:
@RobertImbrie please take a look at my pushed changes and see if they make sense.

mulhod (approved):

RobertImbrie:
@desilinguist Hi! Apologies for the slow response. It does look good, with one exception! On line 754, you change
```
     replace_blanks_with : value, ``dict``, or ``None``, default=None
```
to
```
     replace_blanks_with : value, ``dict``, or ``None``, default=``None``
```
In both sklearn and the previous pushes, the `default=ExampleVariable` section of the parameter header typically doesn't have quotes.

On a related note, did you have any input on using 'value' vs. `Number`? I included a question about it in my initial comment.

desilinguist:
@RobertImbrie good catch, please feel free to fix that quote issue. And yes, number instead of value seems like a better way to describe it! Please go ahead an make that change as well.

RobertImbrie:
@desilinguist Changes have been pushed!

desilinguist (approved):
LGTM! 🎉

PR #702, opened 2021-11-30
Upgrade scikit-learn to v1.0.1

### LINKED ISSUE #699: Update scikit-learn to v1.0

desilinguist:
This PR closes #699. 

This change is pretty straightforward but it’s definitely backwards incompatible which we will reflect in the release version when we put together the release.

The change specifically motivated by the upgrade are:
- Update `requirements.txt` and `conda_requirements.txt` to point to the latest scikit-learn (v1.0.1) and allow up to v1.0.2. 
- Use `squared_error` as the default value of the `loss` parameter for `RANSACRegressor`. 
- Use `toarray()` for converting sparse numpy arrays to dense instead of `todense()` since the latter returns an `np.matrix` instead of an `np.ndarray`. Scikit-learn is planning to drop support for `np.matrix` inputs and is already displaying `FutureWarning`s.
- Force normalize to False for Lars models
  -  Scikit-learn v1.0 will be deprecating the `normalize` attribute for linear models
  - This attribute is set to `False` by default in most scikit-learn linear models anyway and so no warnings are surfaced in SKLL.
  - However, for `Lars`, the default value of `normalize` is still set to `True` and so we need to force it to False to avoid `FutureWarning` instances.
  - The `if` statement added will actually lead to an execption in the `_create_estimator()` method when the `normalize` attribute  doesn't exist, so that will be the perfect reminder to excise it entirely when the time comes.

Other changes include:
- Update Python versions in CI builds (3.9 for Windows/Azure and 3.9 for Linux/Gitlab)
- Update pre-commit and pre-commit hooks to their latest versions

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,7 +4,7 @@ stages:
   - test
 
 variables:
-  PYVERSION: "3.7"
+  PYVERSION: "3.8"
   BINDIR: "/root/sklldev/bin"
   MPLBACKEND: "Agg"
   NOSE_WITH_COV: "1"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.3.0
+    rev: v4.0.1
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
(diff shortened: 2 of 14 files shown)


mulhod:
Should the `scikit-learn` version also be updated in the Conda recipe? It is currently `>=0.24.1,<=0.24.2`. Would this version include `1.0`, e.g. `>=1.0,<=1.0.1`?

desilinguist:
> Should the `scikit-learn` version also be updated in the Conda recipe? It is currently `>=0.24.1,<=0.24.2`. Would this version include `1.0`, e.g. `>=1.0,<=1.0.1`?

Good catch! I will modify it to be the same as the requirements files.

Frost45 (approved):
Looks great 🎉

mulhod (approved):
LGTM

PR #703, opened 2021-12-06
Fix `—drop_blanks` for `filter_features` command line script

### LINKED ISSUE #693: --drop-blanks in filter_features failes if every single row in the feature file has a blank
It's because `pandas` returns an empty data frame and then the DictVectorizer chokes.

desilinguist:
This PR closes #693. 

- Raise an exception if `—drop_blanks` removes all rows in the input file if each of them contains a blank value.
- Add a test for this scenario.
- Update `—help` output for `filter_features` to make it clear that `—drop_blanks` and `—replace_blanks_with` only apply to CSV/TSV files.

### DIFF
--- a/skll/data/readers.py
+++ b/skll/data/readers.py
@@ -364,7 +364,7 @@ class Reader(object):
             # raise an exception here because downstream processing
             # will run into issues
             if df.empty:
-                raise ValueError("No rows/lines left in feature file "
+                raise ValueError("No rows/lines left in the feature file "
                                  "after dropping blank values.")
 
         # if the id column exists,
--- a/skll/data/readers.py
+++ b/skll/data/readers.py
@@ -360,6 +360,13 @@ class Reader(object):
             self.logger.info('Rows/lines with any blank values will be dropped.')
             df = df.dropna().reset_index(drop=True)
 
+            # if the dataframe has no rows left after removing blanks,
+            # raise an exception here because downstream processing
+            # will run into issues
+            if df.empty:
+                raise ValueError("No rows/lines left in feature file "
+                                 "after dropping blank values.")
+
         # if the id column exists,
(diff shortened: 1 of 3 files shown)


desilinguist:
The code coverage issue can be ignored.

skll/data/readers.py
```diff
@@ -360,6 +360,13 @@ def _parse_dataframe(self,
             self.logger.info('Rows/lines with any blank values will be dropped.')
             df = df.dropna().reset_index(drop=True)
 
+            # if the dataframe has no rows left after removing blanks,
+            # raise an exception here because downstream processing
+            # will run into issues
+            if df.empty:
+                raise ValueError("No rows/lines left in feature file "
```
bndgyawali, line 367:
```suggestion
                raise ValueError("No rows/lines left in the feature file "
```

tests/test_commandline_utils.py
```diff
@@ -1835,6 +1835,36 @@ def test_filter_features_with_drop_blanks():
     assert_frame_equal(df_tsv_output, df_expected)
 
 
+@raises(ValueError)
+def test_filter_features_with_drop_blanks_all_blanks():
```
bndgyawali, line 1839: I think this test case will pass when `filter_features.main()` fails on either of the csv/tsv file input. Should we individually check if they both raise ValueError?

mulhod (approved):

tests/test_commandline_utils.py
desilinguist: Good catch! I will add separate tests for CSV/TSV.

bndgyawali (approved):

damien2012eng (approved):

Frost45 (approved):

PR #706, opened 2021-12-08
Rename `pos_label_str` to `pos_label` everywhere.

### LINKED ISSUE #569: Rename `pos_label_str` to `pos_label` and move to Input section
We should implement this in a backwards compatible way along with deprecation warnings in the next release and then fully deprecate the old behavior in the release after that.

desilinguist:
This PR closes #569. 

We are only doing part of the issue here (the renaming) but _not_ moving the `pos_label` field into the `Input` section from the `Tuning` section where it currently lives. This is mainly to minimize the amount of changes that someone would need to go through when converting their old code/config files (renaming can be automated).

### DIFF
--- a/tests/test_classification.py
+++ b/tests/test_classification.py
@@ -71,7 +71,7 @@ def setup():
 def tearDown():
 
     for path in [join(train_dir, 'train_single_file.jsonlines'),
-                 join(train_dir, 'pos_label_str_train.jsonlines'),
+                 join(train_dir, 'pos_label_train.jsonlines'),
                  join(test_dir, 'test_single_file.jsonlines'),
                  join(output_dir, 'rare_class_predictions.tsv'),
                  join(output_dir, 'float_class_predictions.tsv')]:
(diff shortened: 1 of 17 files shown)


tests/test_classification.py
```diff
@@ -227,42 +227,42 @@ def check_label_index_order_with_pos_label_str(use_api=True):
...
     This tests whether binary predictions and probabilities
     obtained from SKLL (via the API as well as the config)
     match the predictions from scikit-learn on the same data
-    for when `pos_label_str` is set and also when it is not set.
+    for when `pos_label` is set and also when it is not set.
     """
 
     if label_list == ['B', 'C']:
-        pos_label_str, other_label_str = 'B', 'C'
+        pos_label, other_label_str = 'B', 'C'
```
Frost45, line 252: I know that this PR is supposed to change only `pos_label_str` to `pos_label` but in a couple of other places, I did see `neg_label_str` being changed to `neg_label`. Similarly, should `other_label_str` be changed too?

Frost45 (approved):

tests/test_classification.py
bndgyawali: I checked out the branch and grepped all the `_label_str` and found the following occurrences of `other_label_str`.
```
./tests/test_classification.py:252:        pos_label, other_label_str = 'B', 'C'
./tests/test_classification.py:254:        pos_label, other_label_str = 2, 1
./tests/test_classification.py:256:        pos_label, other_label_str = 'FRAG', 'NONE'
./tests/test_classification.py:263:        index_label_dict = {0: other_label_str, 1: pos_label}
./tests/test_classification.py:265:        sorted_label_list = sorted([pos_label, other_label_str])
```

bndgyawali (approved):

tests/test_classification.py
desilinguist: Sure, we can do that for consistency. Thanks guys!

PR #707, opened 2021-12-15
Allow specifying custom seed for cross-validation

### LINKED ISSUE #593: Allow the user to control the random seed for cross validation fold generation
Sometimes we want to run the same cross validation experiment multiple times to get an idea of variance across replicates. Currently, it's not easy to manipulate the seed that's used to generate the folds. It's hard coded in [learner.py](https://github.com/EducationalTestingService/skll/blob/master/skll/learner.py#L2363)

It would be nice if this could be a parameter that could be manipulated, either in the config directly, or minimally through the api.

desilinguist:
This PR closes #593. 

- Add a new `cv_seed` keyword argument to the `Learner.cross_validate()` method with the same default value as before.
- Create a new `cv_seed` configuration file field in the “Input” section. Also, show the seed in the experiment log.
- Update the documentation to describe this new field. 
- Add a new input test for "cv_seed" field.
- Add new CV test for the “cv_seed” field
    - Add a new test generator that sets a custom seed and then checks whether the folds generated from inside SKLL match expected folds, for both classifiers and regressors. This also needs two new configuration file templates. 
    - This test requires computing the expected folds which we were already doing for another, already existing test. So, I factored that code out into a utility function in `tests/utils.py`.
- Allow ``cv_seed`` for voting learners too.
    - Add tests for the voting learners side in the same vein as the regular learners using the same utility function.
- Tweak various existing tests due to this extra field. 
- Tweak CI plans for timing
    - `test_voting_learners_api_5.py` now takes 2x longer due to the addition of a new boolean field so let's move it to a different test partition which is much shorter than others.


tests/utils.py
```diff
@@ -412,6 +420,7 @@ def fill_in_config_options_for_voting_learners(learner_type,  # noqa: C901
             param_grid_list,
             sampler_list,
             num_cv_folds,
+            cv_seed,
             learning_curve_cv_folds,
             learning_curve_train_sizes)
```
Frost45, line 426: `create_jsonlines_feature_files()`, `make_classification_data()`, `make_regression_data()` need docstrings.

tests/test_input.py
```diff
@@ -1304,7 +1304,7 @@ def check_cv_folds_and_grid_search_folds(task,
 
 def test_default_number_of_cv_folds():
```
Frost45, line 1306: This file has inconsistent docstrings/summaries for its test functions.

Frost45 (approved):
Looks great @desilinguist! A few minor suggestions.

desilinguist:
@Frost45 please take a look at the changes!

tests/utils.py
```diff
@@ -576,6 +722,44 @@ def make_regression_data(num_examples=100,
...
+        Number of examples in the generated data.
+        Defaults to 100.
+    train_test_ratio : float, optional
+        Ratio of train to test data in the generated data.
+        Defaults to 0.5.
+    num_features : int, optional
+        Number of features in each generated example.
+        Defaults to 2.
+    sd_noise : float, optional
+        Amount of Gaussian noise added to the output
```
Frost45, line 740:
```suggestion
        Amount of Gaussian noise added to the output.
```

Frost45 (approved):
Looks great @desilinguist! Thank you for doing this! 🙏🏼

mulhod (approved):

PR #715, opened 2022-06-28
Added a utility method to the Learner class returning feature names

remo-help:
This PR adds a simple utility method to the Learner class, allowing it to return the feature names used at the estimator step as a 1-D array.
Learners with a selector potentially have a smaller feature name list than is stored in the feature_vectorizer. This method allows for easy access of the actual feature names of features used, even if the selector has removed specific features.

### DIFF
--- a/examples/make_california_example_data.py
+++ b/examples/make_california_example_data.py
@@ -34,7 +34,7 @@ def main():
     X = california.data[chosen_indices, :]
     Y = california.target[chosen_indices]
 
-    # crate example jsonlines dictionaries
+    # create example jsonlines dictionaries
     examples = [
         {"id": f"EXAMPLE_{i}", "y": y, "x": {f"f{j}": x_val for j, x_val in enumerate(x)}}
         for i, (x, y) in enumerate(zip(X, Y))
--- a/skll/data/featureset.py
+++ b/skll/data/featureset.py
@@ -34,7 +34,7 @@ class FeatureSet(object):
     labels : np.array of shape (n_labels,), default=None
         labels for this set.
 
-    feature : list of dict or an array-like of shape (n_samples, n_features), default=None
+    features : list of dict or an array-like of shape (n_samples, n_features), default=None
         The features for each instance represented as either a
         list of dictionaries or an array-like (if ``vectorizer`` is
         also specified).


remo-help:
hold off on review while I add tests

skll/learner/__init__.py
```diff
@@ -634,6 +634,19 @@ def _create_estimator(self):
 
         return estimator, default_param_grid
 
+    def get_feature_names(self):
+        """
+        Feature names of features used by the estimator
+
+        Returns
+        ------
+        1-D array: feature names of features used at the estimator step
```
desilinguist, line 643:
```suggestion
        names : 1-D array
            Names of features actually used by the estimator.
```

skll/learner/__init__.py
```diff
@@ -634,6 +634,19 @@ def _create_estimator(self):
 
         return estimator, default_param_grid
 
+    def get_feature_names(self):
+        """
+        Feature names of features used by the estimator
+
+        Returns
+        ------
```
desilinguist, line 642:
```suggestion
        -------
```

tests/test_output.py
```diff
@@ -1178,3 +1178,34 @@ def test_save_models_to_current_directory():
...
+    """
+    # we create some learners to test on
+    classifier_classes = ['LogisticRegression', 'DummyClassifier', 'LinearSVC']
+    classifiers = [Learner(i) for i in classifier_classes]
+    regressor_classes = ['LinearRegression', 'SVR', 'AdaBoostRegressor']
+    regressors = [Learner(i) for i in regressor_classes]
+
+    # we create some minimal training data for our learners
+    train_fs, test_fs, _ = make_regression_data(num_examples=20,
+                                                  num_features=6)
```
desilinguist, line 1197:
```suggestion
                                                num_features=6)
```

tests/test_output.py
```diff
@@ -1178,3 +1178,34 @@ def test_save_models_to_current_directory():
     eq_(learner1.model_type, learner2.model_type)
     eq_(learner1.model_params, learner2.model_params)
     eq_(learner1.model_kwargs, learner2.model_kwargs)
+
+
+def test_get_feature_names():
```
desilinguist, line 1183: This is a good test but it also needs to cover the case where the feature selector doesn't remove anything. An easy change would be to turn this into a test generator (`check_get_feature_names()`) that takes a boolean argument of whether or not to force the feature selection and then check both cases. Take a look at the other `check_X/test_X` function pairs in `test_output.py`.

desilinguist (dismissed):
Looks pretty good but I have some suggestions - some mostly cosmetic but one to actually increase the test coverage check which is failing right now.

skll/learner/__init__.py
```diff
@@ -634,6 +634,19 @@ def _create_estimator(self):
...
+        1-D array: feature names of features used at the estimator step
+        """
+        if self.feat_selector:
+            return self.feat_vectorizer.get_feature_names_out()[self.feat_selector.get_support()]
+        else:
+            return self.feat_vectorizer.get_feature_names_out()
```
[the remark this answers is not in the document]
remo-help: addressed in #[4f34a5d](https://github.com/EducationalTestingService/skll/commit/4f34a5db7bbe5809760f8937d947c63d0cf44343)

tests/test_output.py
remo-help: addressed in #[4f34a5d](https://github.com/EducationalTestingService/skll/commit/4f34a5db7bbe5809760f8937d947c63d0cf44343)

remo-help (commented):
addressed PR comments

mulhod (approved):

Frost45 (approved):

PR #717, opened 2022-09-09
Replace Boston dataset from examples and tests. 

### LINKED ISSUE #700: Replace boston dataset in tests and examples
Looks like scikit-learn is planning to remove the boston housing dataset in v1.2 because it has ethical issues. They recommend using the calfornia housing dataset instead.  We should do this before the dataset goes away.

Here’s the message you get when using `load_boston()`:

desilinguist:
This PR closes #700.

Replace boston dataset with california dataset everywhere.
- Remove boston example and replace with new california example.
- Remove `make_boston_example_data.py` script and add `make_california_example_data.py` script instead and sub-sample the full dataset is huge.
- Remove references to boston dataset from everywhere.
- Remove boston example test data and add california example test data.
- Replace boston name in sample test data.


doc/run_experiment.rst
```diff
@@ -1075,7 +1075,7 @@ BayesianRidge
 DecisionTreeClassifier and DecisionTreeRegressor
     .. code-block:: python
 
-       {'max_features': ["auto", None]}
+       {'max_features': ["sqrt", None]}
```
mulhod, line 1078: Why was this changed here?

mulhod (approved):

doc/run_experiment.rst
desilinguist: This was a deprecated option that I changed in the code but forgot to change it in the documentation so I included it here.

examples/make_california_example_data.py
```diff
@@ -0,0 +1,58 @@
...
+    print("done")
+
+    # this dataset contains 20,640 samples which is too many
+    # let's just sample this dataset to get 500 samples
+    rng = np.random.default_rng(42)
+    chosen_indices = rng.integers(0, california.target.shape[0], size=500)
+    X = california.data[chosen_indices, :]
+    Y = california.target[chosen_indices]
+
+    # crate example jsonlines dictionaries
```
Frost45, line 37:
```suggestion
    # create example jsonlines dictionaries
```

Frost45 (approved):

examples/make_california_example_data.py
desilinguist: Will fix in the next PR, thanks!

Issue #561, opened 2019-10-16
Add type hints


mulhod:
This will be a pretty large issue. It might make sense to do a module at a time.

PR #729, opened 2023-05-12
Add type hints to `skll.config` module

desilinguist:
- Add `mypy` pre-commit check.
- Add type hints to `skll/config/__init.py` and `skll/config/utils.py`.
- Minor code changes to make improve static typing of the code.
- Update docstrings to use the actual type hints. 

This is the first PR in support of #561, with many more to come.

### DIFF
--- a/skll/config/__init__.py
+++ b/skll/config/__init__.py
@@ -12,6 +12,7 @@ import configparser
 import itertools
 import logging
 import os
+from numbers import Number
 from os.path import basename, join
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Tuple, Union
@@ -20,6 +21,7 @@ import numpy as np
 import ruamel.yaml as yaml
 
 from skll.data.readers import safe_float
+from skll.types import FoldMapping, PathOrStr
 from skll.utils.constants import (
     PROBABILISTIC_METRICS,
     VALID_FEATURE_SCALING_OPTIONS,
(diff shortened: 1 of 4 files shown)


skll/config/__init__.py
```diff
@@ -263,15 +274,65 @@ def validate(self):
...
+    str,
+    bool,
+    Optional[Dict[str, List[str]]],
+    str,
+    str,
+    List[int],
+    List[Number],
+    List[str],
+    bool,
+]:
```
tamarl08, line 327: Wow, that's a long tuple! wouldn't it be better to use Any in this case?
But more importantly, this doesn't match the return value in the docstring below.

skll/config/__init__.py
```diff
@@ -292,7 +353,7 @@ def parse_config_file(config_path, log_level=logging.INFO):  # noqa: C901
     sampler : str
         The name of a sampler to perform non-linear transformations of the input.
 
-    fixed_sampler_parameters : dict
+    fixed_sampler_parameters : List[Dict[str, Any]]
         A dictionary containing parameters you want to have fixed for the sampler.
```
tamarl08, line 357: is it a dictionary or a list of?

skll/config/__init__.py
desilinguist: Lol, yes there must be but I can't really change this right now since it would be a major refactoring.

skll/config/__init__.py
desilinguist: It's a list of dictionaries. I will fix the docstring.

tazin-afrin (approved):

damien2012eng (approved):

PR #731, opened 2023-05-22
Bug fix in feature set split method

### LINKED ISSUE #727: Inconsistent use of ids in FeatureSet.split_by_ids
The method `split_by_ids` receives two sets of ints: `ids_for_split1`, `ids_for_split2`, according to the docstring these are the actual ids (assuming the ids are integers which is not necessarily the case). 

However the method uses `ids_for_split1` as indices of ids in the id list in one case (L519-521), and as the actual ids in the other (L523). `ids_for_split2` is used as indices (L527-529).

https://github.com/EducationalTestingService/skll/blob/1e5d4723f1873a2fa50400845b342ad0d9c6d937/skll/data/featureset.py#L479

tamarl08:
- Rename method from `split_by_ids` to `split` and edit docstring to explicitly say that the split is done on indices and not id values
- Fix the case where split2 is not specified to use indices and not ids
- Add unittests

This PR closes #727.

### DIFF
--- a/skll/data/featureset.py
+++ b/skll/data/featureset.py
@@ -612,7 +612,7 @@ class FeatureSet(object):
 
         # if ids_for_split2 is not given, it will be the complement of ids_split1
         if ids_for_split2 is None:
-            ids_for_split2 = [ind for ind in range(0, len(fs.ids)) if ind not in ids_for_split1]
+            ids_for_split2 = [ind for ind in range(len(fs.ids)) if ind not in ids_for_split1]
 
         ids2 = fs.ids[ids_for_split2]
         labels2 = fs.labels[ids_for_split2] if fs.labels is not None else None
--- a/skll/data/featureset.py
+++ b/skll/data/featureset.py
@@ -564,13 +564,13 @@ class FeatureSet(object):
             return self.ids[value], label, features
 
     @staticmethod
-    def split_by_ids(
+    def split(
         fs: "FeatureSet", ids_for_split1: List[int], ids_for_split2: Optional[List[int]] = None
     ) -> Tuple["FeatureSet", "FeatureSet"]:
         """
         Split ``FeatureSet`` into two new ``FeatureSet`` instances.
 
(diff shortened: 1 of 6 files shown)


skll/data/featureset.py
```diff
@@ -609,16 +609,14 @@ def split_by_ids(
...
-            ids2 = fs.ids[~np.in1d(fs.ids, ids_for_split1)]
-            labels2 = fs.labels[~np.in1d(fs.ids, ids_for_split1)] if fs.labels is not None else None
-            features2 = (
-                fs.features[~np.in1d(fs.ids, ids_for_split1)] if fs.features is not None else None
-            )
-        else:
-            ids2 = fs.ids[ids_for_split2]
-            labels2 = fs.labels[ids_for_split2] if fs.labels is not None else None
-            features2 = fs.features[ids_for_split2] if fs.features is not None else None
+            ids_for_split2 = [ind for ind in range(0, len(fs.ids)) if ind not in ids_for_split1]
```
desilinguist, line 615:
```suggestion
            ids_for_split2 = [ind for ind in range(len(fs.ids)) if ind not in ids_for_split1]
```

tests/test_featureset.py
```diff
@@ -1228,3 +1228,63 @@ def test_drop_blanks_and_replace_blanks_with_raises_error():
     csv_path = other_dir / "test_drop_blanks_error.csv"
     _create_test_file(csv_path, test_csv)
     CSVReader(csv_path, replace_blanks_with=4.5, drop_blanks=True).read()
+
+
+def test_split_two_id_sets():
+    """Test split_by_ids with two input id sets."""
```
desilinguist, line 1234:
```suggestion
    """Test split() with two input id sets."""
```

desilinguist (changes requested):
Just a couple of minor tweaks. Looks great! Thank you for adding the comprehensive tests!!

desilinguist (approved):

PR #738, opened 2023-06-12
Improve docstrings and type hints (Part 2). 

desilinguist:
- chore(req): revert sphinx to <6.0. The sidebar does not work properly with sphinx>=6.0.
- chore: ignore E501 errors in `featureset.py`.
- chore: fix docstrings and types in `skll.data`.  Also refactored the keyword arguments for many writers to be more explicit rather than using `**kwargs` everywhere.
- docs: improve layout of `api/data.rst`. Use individual `autoclass` items instead of one `automodule` item to make sure  that `Reader` and `Writer` base classes are shown before all the other classes.

To review this, please look at the [built documentation](https://skll.readthedocs.io/en/647-improve-docstrings-typehints-2/index.html) and check that under the "API documentation" > "`data` Package`":

- All parameters have a docstring. 
- All type hints  have clickable types or type hints. 
- The default values are all specified as `default = blah`. 
- All other changes look okay. 

This is part of #647.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -155,6 +155,6 @@ README file in the doc/ directory for more information.
 
 For building the documentation, you will need [sphinx](http://sphinx.pocoo.org/) as well as the readthedocs sphinx theme. To install both, just run:
 
-      $ conda install 'sphinx>=6,<7' sphinx_rtd_theme==1.2.0
+      $ conda install 'sphinx<6' sphinx_rtd_theme==1.2.0
 
 in your existing conda environment.
--- a/skll/data/writers.py
+++ b/skll/data/writers.py
@@ -28,11 +28,14 @@ from skll.types import FeatGenerator, FeatureDict, IdType, LabelType, PathOrStr
 
 class Writer(object):
     """
-    Helper class for writing out FeatureSets to files on disk.
+    Write out FeatureSets to files on disk.
+
+    This is the base class used to create featureset writers for different
+    file types.
 
     Parameters
     ----------
-    path : PathOrStr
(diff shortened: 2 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -155,6 +155,6 @@ README file in the doc/ directory for more information.
 
 For building the documentation, you will need [sphinx](http://sphinx.pocoo.org/) as well as the readthedocs sphinx theme. To install both, just run:
 
-      $ conda install 'sphinx>=6,<7' sphinx_rtd_theme==1.2.0
+      $ conda install 'sphinx<6' sphinx_rtd_theme==1.2.0
```
dblandan, line 158: Interesting...  What happened that made us need to roll back to an earlier version of sphinx?

desilinguist: The sidebar was no longer expanding correctly with the nice + signs etc.

skll/data/writers.py
```diff
@@ -860,14 +1020,15 @@ def _write_line(
...
-                f"{self._sanitize(self.label_map[label_])}=" f"{self._sanitize(label_)}",
-                end=" | ",
-                file=output_file,
-            )
-        else:
-            print(" |", end=" ", file=output_file)
+        if self.label_map:
+            if label_ in self.label_map:
+                print(
+                    f"{self._sanitize(self.label_map[label_])}=" f"{self._sanitize(label_)}",
```
dblandan, line 1026:
```suggestion
                    f"{self._sanitize(self.label_map[label_])}={self._sanitize(label_)}",
```

dblandan (approved):

Frost45 (approved):

PR #742, opened 2023-06-20
Add BaggingClassifier and BaggingRegressor support

### LINKED ISSUE #614: Add BaggingClassifer and BaggingRegressor
Should be fairly easy since it's pretty much like AdaBoost in terms of intergration.

desilinguist:
- Fix: base estimators cannot be serialized
  - It turns out that there's a long standing bug in SKLL where using meta-estimators like AdaBoost* that use a base estimator are unable to be used with `run_experiment` since their base estimator prevents the model parameters dictionary from being serialized to disk. Added code that replaces the base estimator object with a string before serializing.
  - Existing tests did not cover this scenario since they used those meta-estimators via the API and not via a configuration file. Added new tests that covered this scenario.
- Add support for `BaggingClassifier` and `BaggingRegressor`
  - Enable them to be used in SKLL. 
  - Set default parameter grids for grid search. 
- Add documetnation for Bagging{Classifier,Regressor} use.
- New tests for Bagging{Classifier,Regressor}

Easiest way to review is to try out the new learners (along with different `estimator` fixed parameter values) in the California and Titanic examples. 

This PR closes #614.


tests/test_classification.py
```diff
@@ -846,6 +848,41 @@ def test_adaboost_predict():  # noqa: D103
...
+    base_estimator : str
+        Name of base estimator to use.
+    oob_score : bool
+        Whether to use out-of-bag samples to estimate generalization error.
+    expected_score : float
+        Expected score for the predictions.
+    """
+    train_fs, test_fs = make_sparse_data()
+
+    # train an AdaBoostClassifier on the training data and evalute on the
```
Frost45, line 866:
```suggestion
    # train an AdaBoostClassifier on the training data and evaluate on the
```

tests/test_regression.py
```diff
@@ -507,11 +537,39 @@ def check_ransac_regression(base_estimator, pearson_value):
...
-        [None, "SGDRegressor", "DecisionTreeRegressor", "SVR"], [0.95, 0.45, 0.75, 0.64]
+        [None, "SGDRegressor", "DecisionTreeRegressor", "SVR"], [0.95, 0.95, 0.75, 0.64]
     ):
         yield check_ransac_regression, base_estimator_name, pearson_value
 
 
+def check_bagging_regression(base_estimator, pearson_value):
+    train_fs, test_fs, _ = make_regression_data(num_examples=2000, sd_noise=4, num_features=3)
+
+    # train a RANSACRegressor on the training data and evalute on the
```
Frost45, line 548:
```suggestion
    # train a RANSACRegressor on the training data and evaluate on the
```

Frost45 (commented):
Testing it out locally!

Frost45 (approved):
Tested it out with Titanic! Everything works as expected 🎉

PR #743, opened 2023-06-22
Add support for `HistGradientBoostingClassifier` and `HistGradientBoostingRegressor`

### LINKED ISSUE #489: Add HistGradientBoostingClassifier and HistGradientBoostingRegressor
There is a [new experimental implementation](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingClassifier.html#sklearn.ensemble.HistGradientBoostingClassifier) for GBTs which is supposed to be orders of magnitude faster than the vanilla GBTs for N > O(10K) which is not an uncommon use case for SKLL users. 

We should include these once they move from experimental to stable versions. Not now.

desilinguist:
- Add support for HistGradientBoosting learners
    - Add imports.
    - Add default parameter grids.
- Update documentation:  
    - Document default fixed parameters and parameter grids.
    - Add the same for Bagging learner that we forgot to do before.
    - Also add explicit notes to all learners that require dense features.
    - Update ransac regressor note to mention base estimators explicitly.
- Update regression and classification tests. 

Again, best way to to test is to try out the new learners with the California and Iris examples. 

This PR closes #489.


dblandan (approved):

mulhod (approved):
Everything worked perfectly. Ran on M1 FYI.

Ran experiments on California and Iris.

Frost45 (approved):
Tested locally! Everything works 🎉

Issue #744, opened 2023-06-22
Add support for Python 3.11


desilinguist:
Make sure SKLL works with Python 3.11.

desilinguist:
Turns out that `nose` no longer works with python 3.11. So, adding support for Python 3.11 means that we will need to replace `nose` with either `nose2` or `pytest`.

desilinguist:
This is blocked by #746.

PR #747, opened 2023-06-30
Migrate tests to `nose2` instead of `nose`

### LINKED ISSUE #746: Migrate from `nose` to `nose2` for testing
`nose` no longer works with Python 3.11 so we need to move to `nose2` if we want to support Python 3.11

desilinguist:
- Update all tests to be nose2-compatible. 
- Update pre-commit config to ignore docstyle and flake8 errors in test files. 
- Replace `nose` with `nose2` in requirements file.
- Update Gitlab and Azure CI builds to use nose2. 
- Add `unittest.cfg` for nose2 configuration 
- Move testing utilities into new `skll.utils.testing` module since `nose2` doesn't pick up the `tests` directory as a module. 
- Update all tests to use `skll.utils.testing`. 
- Update `.coveragerc` with proper exclusions. 
- Disable logging in very noisy test file.

This PR closes #746.


mulhod:
There are still references to `nosetests` in `CONTRIBUTING.md`. Also it is said there that `python=3.10` is required for setting up a development environment. Maybe that's still true for other reasons, but I think it should be considered. I actually set up an environment that uses 3.11 and I think the unit tests are running fine with `nose2`.

desilinguist:
> There are still references to `nosetests` in `CONTRIBUTING.md`. Also it is said there that `python=3.10` is required for setting up a development environment. Maybe that's still true for other reasons, but I think it should be considered. I actually set up an environment that uses 3.11 and I think the unit tests are running fine with `nose2`.

Good catch on `CONTRIBUTING.md`. As for Python 3.11, there's a separate issue #744 for that which I will work on after this one is merged. As part of that, I will be switching the Azure build to 3.11 from 3.9.

mulhod:
All unit tests passed with 3.11.

mulhod (approved):

skll/utils/testing.py
dblandan, line 1: I noticed that the methods with returns don't have type hints for the return values...  I'm not 100% sure, but I think the other recent PRs included these.  (e.g., the `def` for `fill_in_config_paths(...)` should probably include `-> Path`.

tests/test_classification.py
```diff
@@ -47,157 +47,266 @@
...
+        self.assertEqual(contiguous_ints_or_floats([1, 2, 3, 4.0]), True)
+        self.assertEqual(contiguous_ints_or_floats([-1, 1]), False)
+        self.assertEqual(contiguous_ints_or_floats([2, 4, 6]), False)
+        self.assertEqual(contiguous_ints_or_floats([3, 6, 11]), False)
+        self.assertEqual(contiguous_ints_or_floats([-2.0, -1.0, 1.0, 2.0]), False)
+        self.assertEqual(contiguous_ints_or_floats([1.0, 1.1, 1.2]), False)
+        self.assertRaises(TypeError, contiguous_ints_or_floats, ["a", "b", "c"])
+        self.assertRaises(TypeError, contiguous_ints_or_floats, np.array([1, 2, 3, "a"]))
+        self.assertRaises(ValueError, contiguous_ints_or_floats, [])
+        self.assertRaises(ValueError, contiguous_ints_or_floats, np.array([]))
```
dblandan, line 130: Not critical, but I'd prefer to see separate tests for error handling.

desilinguist: Care to elaborate?

tests/test_classification.py
```diff
@@ -209,1731 +318,1591 @@ def check_label_index_order_with_pos_label(use_api=True):
...
-        # (d) probabilistic metrics and no probabilities
-        skipped_conditions = (
-            (metric in ["average_precision", "roc_auc"] and len(label_array) != 2)
-            or (
-                (metric in WEIGHTED_KAPPA_METRICS or metric in CORRELATION_METRICS)
-                and issubclass(label_array.dtype.type, str)
-            )
-            or (metric in WEIGHTED_KAPPA_METRICS and not contiguous_ints_or_floats(label_array))
-            or (metric in PROBABILISTIC_METRICS and not use_probabilities)
+                # 5. The only ones left are the weighted kapps;
```
dblandan, line 1360:
```suggestion
                # 5. The only ones left are the weighted kappas;
```

tests/test_commandline_utils.py
dblandan, line 1: Another general observation: I really appreciate the effort & scope of this refactor! Will there be variable type hints added later? There are a few places where I think they would be helpful, but I don't think it's worth holding up this PR for them...

desilinguist: We aren't adding type hints to test files.

tests/test_input.py
```diff
@@ -10,2690 +10,2624 @@
...
+            _,
+            _,
+            _,
+            _,
+            _,
+            _,
+            _,
+            output_metrics,
+            _,
+        ) = parse_config_file(config_path)
```
dblandan, line 885: ok... i know this is from the pre-commit, but yuck...  lol

dblandan: i feel like it might actually be more readable for these to do something like:
```
ph = parse_config_file(config_path)
do_grid_search = ph[14]
grid_objectives = ph[15]
output_metrics = ph[-2]
```

tests/test_input.py
```diff
@@ -10,2690 +10,2624 @@
...
+            _,
+            _,
+            _,
+            _,
+            _,
+            _,
+            _,
+            _,
+            _,
+        ) = parse_config_file(config_path)
```
dblandan, line 959: Especially in cases where you only need one value...
```suggestion
        train_path = parse_config_file(config_path)[37]
```

dblandan (approved):

tests/test_input.py
desilinguist: Good call! Definitely less yucky :)

PR #766, opened 2024-02-16
Update scikit-learn to v1.4

desilinguist:
- Update sklearn version in `requirements.txt`. 
- Properly clean up temporary files in `test_output`. 
- Update test data due to sklearn change.
- Add filter for unnecessary `matplotlib` messages.
- Fix warnings and filter `matplotlib` messages.
  - Remove deprecated `scale` parameter from seaborn calls.
  - Add explicit `palette` call when plotting time curve since that's now required.
  - Use `matplotlib.category` msg filter to filter out unnecessary msgs.
- Update pre-commit checks.
- Handle sklearn deprecations.
    - `needs_proba` and `needs_threshold` arguments for `make_scorer()` are deprecated in favor of `response_method`.
    - `LinearSVC` and `LinearSVR` now need to have `dual` explicitly set to "auto".
    - `AdaBoostClassifier` now needs to have algorithm` set to "SAMME".
    - Update custom metrics documentation.
    - Update tests and test data.
- Update Gitlab CI python version to 3.10 since sklearn v1.4 drops support for both Python 3.8 and 3.9.


tamarl08 (approved):

doc/custom_metrics.rst
```diff
@@ -12,15 +12,27 @@ Writing Custom Metric Functions
...
+   - If the value is ``None``, it is the same as ``"predict"``.
+
+   The default value for ``response_method`` is ``None``.
 
 Note that these keyword arguments are identical to the keyword arguments for the `sklearn.metrics.make_scorer() <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.make_scorer.html#sklearn.metrics.make_scorer>`_ function and serve the same purpose.
 
-In short, custom metric functions take two required positional arguments (order matters) and three optional keyword arguments. Here's a simple example of a custom metric function: F\ :sub:`β` with β=0.75 defined in a file called ``custom.py``.
+.. important::
+
+   Previous versions of SKLL offered the ``needs_proba`` and ``needs_threshold`` keyword arguments for custom metrics but these are now deprecated in scikit-learn and replaced by the ``response_method`` keyword argument. To replicate the behavior of ``needs_proba=True``, use ``response_method="predict_proba"`` instead and to replicate ``needs_threshold=True``, use ``response_method=("decision_function", "predict_proba")`` instead.
```
mulhod, line 33:
```
To replicate the behavior of ``needs_proba=True``, use ``response_method="predict_proba"`` instead and to replicate ``needs_threshold=True``, use ``response_method=("decision_function", "predict_proba")`` instead.
```

Is the second `response_method` value correct? I was expecting it to look like the first but with `needs_threshold` in it somewhere (not `predict_proba`). It also is a couple strings in parens whereas the first was just a string.

mulhod (approved):

PR #769, opened 2024-03-08
v5.0.1 Release

desilinguist:
- Update pre-commit checks.
- Update dependencies.
  - Remove all dev dependencies from `requirements.txt`.
  - Update versions in `doc/requirements.txt`.
- Update version numbers in `version.py` and in the conda recipe.
- Add new `requirements.dev` file. This file contains the runtime as well as dev dependencies.
  - Update `CONTRIBUTING.md` to use this file instead of `requirements.txt`.
  - Exclude this file in `MANIFEST.in` so that it's not part of the PyPI package.
  - Update CI pipelines to use `requirements.dev` instead of `requirements.txt`
- Update release process checklist.

✅ Conda package was created successfully, [uploaded](https://anaconda.org/ets/skll) to anaconda.org, and [tested](https://github.com/EducationalTestingService/skll-conda-tester/pull/11).
✅ PyPI package was created successfully, [uploaded](https://test.pypi.org/project/skll/) to TestPyPI, and [tested](https://github.com/EducationalTestingService/skll-pip-tester/pull/11).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ How to contribute
 
 3. Create an isolated environment for SKLL development. We recommend using the [conda](https://conda.io/en/latest/) package manager. To create a `conda` environment, run the following command in the root of the working directory:
 
-         $ conda create -n sklldev -c conda-forge --file requirements.txt python=3.11
+         $ conda create -n sklldev -c conda-forge --file requirements.dev python=3.11
 
 4. Activate the conda environment
 
@@ -52,13 +52,13 @@ How to contribute
 
          $ pre-commit run
 
-   To run the `isort` hook alone on changed files:
+   To run the `black` hook alone on changed files:
 
-         $ pre-commit run isort
+         $ pre-commit run black
 
    To run the `isort` hook alone on a given file:
 
-         $ pre-commit run isort <file-path>
+         $ pre-commit run black <file-path>
 
    Finally, the `SKIP` environment variable can be used to indicate to
    `pre-commit` that certain checks should be skipped. It can be
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,7 +56,7 @@ How to contribute
 
          $ pre-commit run black
 
-   To run the `isort` hook alone on a given file:
+   To run the `black` hook alone on a given file:
 
          $ pre-commit run black <file-path>
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,14 +1,14 @@
 [tool.black]
 include = '\.pyi?$'
 line-length = 100
-target-version = ['py38']
+target-version = ['py311']
 
 [tool.ruff]
 extend-exclude = ["setup.py"]
 lint.select = ["D", "E", "F", "I"]
 lint.ignore = ["D212"]
 line-length = 100
-target-version = "py38"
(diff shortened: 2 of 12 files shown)


requirements.dev
```diff
@@ -0,0 +1,16 @@
+beautifulsoup4
```
tamarl08, line 1: remove beautifulsoup?

desilinguist: Actually, turns out we do still use it in `readers.py`. Sorry!

twoertwein: Based on the type annotations, it cannot be bytes `file: IO[str]` :)
https://github.com/EducationalTestingService/skll/blob/affb97a658e9f9e679ada0fe530f9dccb84f9b5a/skll/data/readers.py#L868

I assume `bs4.UnicodeDammit` cleverly guesses the encoding(?) - could maybe have a keyword argument for the encoding to get rid of bs4.

desilinguist: Yep, we used it for converting to the encoding we needed since we used to deal with a lot of weirdly encoded files back in the day. May be we don't need it anymore but I don't want to investigate that in this MR.

CONTRIBUTING.md
```diff
@@ -52,13 +52,13 @@ How to contribute
...
-   To run the `isort` hook alone on changed files:
+   To run the `black` hook alone on changed files:
 
-         $ pre-commit run isort
+         $ pre-commit run black
 
    To run the `isort` hook alone on a given file:
 
-         $ pre-commit run isort <file-path>
+         $ pre-commit run black <file-path>
```
twoertwein, line 61: do you mean ruff? `--select "D,E,F,I"` will sort imports (if you let ruff `--fix` it)

there is also ruff format now - could remove black:

```
- repo: https://github.com/astral-sh/ruff-pre-commit
  # Ruff version.
  rev: v0.3.1
  hooks:
    # Run the linter.
    - id: ruff
      args: [ --fix ]
    # Run the formatter.
    - id: ruff-format
```

desilinguist: Yes, we are slowly moving towards ruff and may be we can use it to get rid of black too but I'd like to test it a bit first to make sure that it has full parity. I'll file an issue.

desilinguist: BTW, we already have `ruff` in the pre-commit checks which is why we removed `isort`.

tamarl08 (approved):

CONTRIBUTING.md
twoertwein: Should also be `black` on line 59

twoertwein (approved):

pyproject.toml
```diff
@@ -1,14 +1,14 @@
...
-target-version = ['py38']
+target-version = ['py311']
 
 [tool.ruff]
 extend-exclude = ["setup.py"]
 lint.select = ["D", "E", "F", "I"]
 lint.ignore = ["D212"]
 line-length = 100
-target-version = "py38"
+target-version = "py311"
```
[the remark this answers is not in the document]
desilinguist: The minimum supported version for development at ETS is indeed 3.11 even though we technically support 3.10 of people want to run SKLL with it.

twoertwein (approved):

mulhod (approved):

PR #773, opened 2024-07-25
Replace setup.py with pyproject.toml

tamarl08:
- [x] TODO: update documentation

See the package in [testPyPI](https://test.pypi.org/project/skll/5.0.2/) (I incremented the version just for the test)


pyproject.toml
```diff
@@ -1,3 +1,68 @@
...
+filter_features = "skll.utils.commandline.filter_features:main"
+generate_predictions = "skll.utils.commandline.generate_predictions:main"
+join_features = "skll.utils.commandline.join_features:main"
+print_model_weights = "skll.utils.commandline.print_model_weights:main"
+run_experiment = "skll.utils.commandline.run_experiment:main"
+skll_convert = "skll.utils.commandline.skll_convert:main"
+summarize_results = "skll.utils.commandline.summarize_results:main"
+compute_eval_from_predictions = "skll.utils.commandline.compute_eval_from_predictions:main"
+plot_learning_curves = "skll.utils.commandline.plot_learning_curves:main"
+
```
desilinguist, line 56:
```
[tool.setuptools]
packages = ["skll"]
```

desilinguist (changes requested):

desilinguist:
I tested both `pip install -e .` and `python -m build` and both work fine. I suggested to add a section since I had it in my other `pyproject.toml` files and figured it can't hurt.

pyproject.toml
tamarl08: isn't it redundance to have it with the ```[tool.setuptools.packages.find]``` part?
[This](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#custom-discovery) is the documentation I'm looking at.
Maybe I shouldn't set `namespace` to false?

desilinguist: Yeah looks like `python -m build` finds all the correct files so may be we don't need this. Can you try setting up an environment with the new package from TestPyPI to make sure it works?

tamarl08: that's what I did. you mean with this change?

desilinguist (approved):

damien2012eng (approved):

PR #775, opened 2024-09-09
Allow None value in feature_scaling

tamarl08:
If `None` is given in the configuration file, it is passed as the string "None". This change converts it to "none" which is the default value.
I didn't change anything in the documentation - @desilinguist  do you `None` this should be explicitly mentioned as an acceptable option?

### DIFF
--- a/skll/config/__init__.py
+++ b/skll/config/__init__.py
@@ -720,8 +720,8 @@ def parse_config_file(
 
     # ensure that feature_scaling is specified only as one of the
     # four available choices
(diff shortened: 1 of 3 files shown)


desilinguist:
> I didn't change anything in the documentation - @desilinguist do you `None` this should be explicitly mentioned as an acceptable option?

@tamarl08 yes I think we should. Not only that `None` is now an alias for "none"/"None" but that other options are also case-insensitive.

desilinguist (approved):
LGTM! Thanks!

damien2012eng (approved):

PR #777, opened 2024-10-19
Update scikit learn to the latest version

damien2012eng:


desilinguist:
We usually keep the pin but just increment the versions on both ends since sometimes bigger versions changes break things.

desilinguist (approved):
LGTM!

Frost45 (approved):

